PR #35, opened 2013-09-12
Add option to hide Welcome screen

ViViDboarder:
The welcome screen doesn't provide any options over what is available on
the main window. In fact it provides less options as the user cannot
access Settings while the welcome screen is open.

This commit adds a new settings option for hiding the welcome screen as
well as a section in settings for GUI or Loader related settings
independent of the API.

Welcome Screen defaults to being shown


diracz:
Sorry about the delay. This change looks good. In fact we might consider setting the default to hide the welcome window in the future.

PR #36, opened 2013-09-12
Provide access to endpoint during login

ViViDboarder:
In order to change your login to a sandbox from production once
selecting an action required the user to cancel, go to settings, change
the endpoint, save, and finally go back into the action.

This patch provides a textbox for changing the login url straight from
the action settings page


ViViDboarder:
That makes sense. Just moving the field outside of the if block should make it available for both types of login flows.

diracz:
ViViDboarder, I'm going to close the issue here. This request is really about adding extra text input field. 
If for some reason the sfdcInternal option is not working for you, we can consider separating the url field with sfdcInternal=true.

PR #40, opened 2014-09-03
support ${} syntax. it's provide to access environment.

megascus:
see
http://stackoverflow.com/questions/18744663/how-to-inject-environmental-variables-inside-spring-xml-configuration

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -122,19 +122,8 @@
 
     <dependency>
       <groupId>org.springframework</groupId>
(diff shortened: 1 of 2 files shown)


tggagne:
Should pull request #40 have allowed me to use ${DATE} in process-conf.xml as in

<entry key="process.outputSuccess" value="log/success/importService.csv.${DATE}" />

assuming DATE is defined in my environment?
```
DrozBook:sql tgagne$ echo $DATE
20150417
```

PR #42, opened 2014-09-26
Tabs and commas separate fields, fix regression.

gnguyen-sfdc:
Use CSVReader update to separate fields on commas or tabs.


gnguyen-sfdc:
No idea. I honestly thought it was, but I guess I was mistaken.

PR #63, opened 2015-04-14
undo pull request #40

tggagne:
This request undoes pull #40 , and it works as advertised (no file-not-found error).  But I think the better option might be to simply add "file://" service identifier to the current code.

After forking, I executed a 

1. git revert -m 1 a5a188f7f4b32c02f291acb9c4ad5dbb805f5d33
2. fixed-up pom.xml
3. committed to tggagne/dataloader and 
4. created the pull request.

Hopefully, these were the right steps.  That's more git surgery than I'm accustomed to.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -122,15 +122,23 @@
 
     <dependency>
       <groupId>org.springframework</groupId>
-      <artifactId>spring-context</artifactId>
+      <artifactId>spring-beans</artifactId>
       <version>${spring.version}</version>
     </dependency>
+
+    <dependency>
+      <groupId>org.springframework</groupId>
+      <artifactId>spring-core</artifactId>
+      <version>${spring.version}</version>
+    </dependency>
+
     <dependency>
       <groupId>org.springframework</groupId>
       <artifactId>spring-jdbc</artifactId>
       <version>${spring.version}</version>
       <scope>test</scope>
     </dependency>
+
     <dependency>
       <groupId>org.testng</groupId>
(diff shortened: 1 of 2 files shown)


diracz:
What's what I was suspecting for to fix your file path.

tggagne:
Sorry about that.  I thought the request to, "Can you revert that change and see if that will work?" meant to undo #40.

It looks like you already merged it into master.  How do you want to proceed from here?

How about you undo the merge, I'll fork another copy and add "file://" as I suggested earlier, and submit a new pull request?

diracz:
Sorry I'm a bit confused. Which of the following is correct?
1. Do you know undoing #40 will fix the file path problem, without your "file://" fix? This is the first question I'd like to know. If yes, this merge has already been in master, and it fixes your problem. 
2. If 1) won't work, and adding "file://" works WITHOUT undoing #40 works and fixes the problem, I'll revert this merge, and you can submit another pull request. 
- If 1) won't work, and adding "file://" works WITH undoing #40 works, then I'll not revert, and you submit another pull request for the "file://" change.

tggagne:
Option 1 fixes the problem, but at the cost of losing the feature of 
substituting environment variables inside the conf files added by #40.

Adding "file://" to the code also fixes the problem and preserves the 
feature, so I think reverting the merge is the correct action, and I'll 
submit a new pull request.

diracz:
Thanks for confirming. 
#65 just reverted it. Go ahead submitting the "file://" fix.

Issue #59, opened 2015-04-08
Upsert from database takes /forever/


tggagne:
I'm investigating why moving LOTS of rows from a database into Salesforce takes so long.  Basically, the process reports "Loading: upsert" then does nothing for hours. Nothing is written to the output.  No batches are created.  Nothing is sent to Salesforce.

I suspect the entire result-set is being read before anything is loaded to Salesforce.  If one wants to migrate a few million rows this is basically useless.

Has anyone else tripped over this?

I'm looking at DatabaseReader.java, line 145 which reads:
// execute the query and save the result set
dbContext.setDataResultSet(statement.executeQuery());

which looks like it plans to spool all the results before doing anything.  There doesn't seem to be any looping over the rows directly into Salesforce.

tggagne:
It actually looks like dataloader first counts all the rows in the source data before moving them.  At least, I out a logger.error() inside DatabaseReader.readRow() and for a 39-row table it was called 81 times.

If this is the case, I'll be trying to figure out if that is a /requirement/ or something that can be skipped.

tggagne:
I'm unsure what the consequences will be, but this is my minor change which seems to be working.  I'll try loading a few hundred thousand rows tonight to see what happens.

```
diff --git a/src/main/java/com/salesforce/dataloader/dao/database/DatabaseReader.java b/src/main/java/com/salesforce/dataloader/dao/database/DatabaseReader.java
index c1f2d40..5bfde44 100644
--- a/src/main/java/com/salesforce/dataloader/dao/database/DatabaseReader.java
+++ b/src/main/java/com/salesforce/dataloader/dao/database/DatabaseReader.java
@@ -210,10 +210,7 @@ public class DatabaseReader implements DataReader {
 
     @Override
     public int getTotalRows() throws DataAccessObjectException {
-        if (totalRows == 0) {
-            totalRows = DAORowUtil.calculateTotalRows(this);
-        }
-        return totalRows;
+        return 0;
     }
```

diracz:
As a quick general note, data loader supports up to 5 million records. If your "LOTS" of rows is bigger than that, there's a chance that it won't work.

tggagne:
diracz, whom do I contact to become a collaborator so I may push a branch and create a pull request?

diracz:
[...]

Issue #62, opened 2015-04-13
LoadRateCalculator -- does it work?


tggagne:
Even though I zeroed-out the number of records to process value to patch #59, I'm bothered that the messages for records-per-second is so off.  At one point during a batch upload it was reporting 600 million records/minute.

I suspect it's dividing the whole of the records processed so far by the duration of the most recent batch--so each subsequent batch reports higher-and-higher rates.

Has anyone else noticed this? Has anyone else looked into the *RateCalculator classes?

bigassforce:
We get negative numbers on big loads @tggagne

> [java] 2017-05-20 23:21:23,448 INFO  [execute] progress.NihilistProgressAdapter setSubTask (NihilistProgressAdapter.java:78) - Processed 575,600 of 0 total records.
> Rate: 3,341,109,000 records per hour. Estimated time to complete: 0 minutes and 0 seconds.
> Rate: 24,804,000 records per hour. Estimated time to complete: 0 minutes and 0 seconds.
> Rate: -3,581,000 records per hour. Estimated time to complete: 0 minutes and 0 seconds.
> There are 705,156 successes and 3,444 errors.

ashitsalesforce:
It has been a while since the issue was reported. A code review of the current code in LoadRateCalculator.java (calculateSubTask) does not indicate how a negative rate gets generated. Moreover, log messages showing rate are no longer logged, perhaps due to code changes since the issue was filed. Closing it for now. Reopen or report a new one if the issue is still seen.

PR #67, opened 2015-04-15
Fixes issue #59, but makes #62 even more annoying than before.

tggagne:
This is the only way dataloader works well with 4+ million rows in the source data.  I tested it using an Oracle PS source database to migrate records to Salesforce.  Rather than dataloader appearing to do nothing for over an hour (as it read rows from Oracle), it now starts creating batches within a few minutes.

### DIFF
--- a/src/main/java/com/salesforce/dataloader/dao/database/DatabaseReader.java
+++ b/src/main/java/com/salesforce/dataloader/dao/database/DatabaseReader.java
@@ -210,10 +210,7 @@ public class DatabaseReader implements DataReader {
 
     @Override
     public int getTotalRows() throws DataAccessObjectException {
-        if (totalRows == 0) {
-            totalRows = DAORowUtil.calculateTotalRows(this);
-        }
-        return totalRows;
+        return 0;
     }
 
     @Override


diracz:
I don't feel comfortable merging your change here. 
I understand that getTotalRows() is not gonna affect anything functionally, since it's only for logging and monitoring purpose. But just blinding returning zero is a bit too blunt, and may create confusions later on. 
What I suggest: create a config parameter, to control whether we want to keep track of the number of records from DAO. Defaulting it to true, and you can change it to false locally.

tggagne:
Do you (or does anyone else) have a recommendation on where that config 
parameter would go or what it should be called?

I'm thinking it should be something like:

 1. dataAccess.fetchRowsTwice (default=false) or
 2. dataAccess.skipTotalCount (default=true) or
 3. dataAccess.quickRead (default=true)

or something.  I'm open to suggestions.

Hey, who updates the documentation?

diracz:
skipTotalCount sounds good. 
Salesforce updates the "official" documentation. I can pass the new parameter to the doc team. But for some configs like this, adding some inline documentation is good enough for now.

PR #69, opened 2015-04-16
Created new property, daoAccess.skipTotalCount

tggagne:
This is a better solution that was submitted earlier today.  Rather than just skipping the row count, the property allows users to decide whether they want it calculated and reported in progress updates or not.

The default value is true.

### DIFF
--- a/src/main/java/com/salesforce/dataloader/config/Config.java
+++ b/src/main/java/com/salesforce/dataloader/config/Config.java
@@ -83,6 +83,14 @@ public class Config {
     public static final int DEFAULT_BULK_API_BATCH_SIZE = 2000;
     public static final long DEFAULT_BULK_API_CHECK_STATUS_INTERVAL = 5000L;
 	public static final String DEFAULT_ENDPOINT_URL = "https://login.salesforce.com";
+	
+	/*
+	 * Issue #59 - Dataloader will not read all the database rows to get a total count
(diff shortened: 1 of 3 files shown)


diracz:
Looks good, except that you included the DataSource class in this pull. Can you remove that file, so it can get merged?

tggagne:
Hmm.  I didn't push DataSource until today, and the pull request was 
from yesterday or this AM, wasn't it?

So, if I need to remove it are you suggesting I revert the commit? Sorry 
for the newbie questions--I'm not yet confident in my fork/pull moves.

tggagne:
I'm looking again at the pull request -- that was submitted 23 hours 
ago.  Why are commits I made this afternoon included in it?

Can I submit another pull request from the commit before DataSource?

diracz:
I'm not an expert on github by any means, but I think the basic idea is to folk per feature, so that you don't get into multiple features on the same fork like here. 
You can either revert the new file (of course, create another fork and put the new file there for the encrypt password feature), or do something like this to split commit:
http://stackoverflow.com/questions/12660839/pull-request-for-only-certain-files-commits

tggagne:
Ok.  This pull request can be killed.  I've submitted request #70.

tggagne:
I think I figured it out.  See Pull request #70. I closed #69.

Issue #64, opened 2015-04-15
Can the database password be encrypted in database-conf.xml?


tggagne:
I want to push my database-conf.xml file to a git repository, but am uncomfortable doing so with my password in the clear.

I haven't read in the documentation that DB passwords can be encrypted like the SFDC passwords can.

Does anyone know?

tggagne:
Apparently, the same "encrypt" command used for obscuring passwords in config.properties works for database-conf.xml, too.

tggagne:
I should re-open this.  The encrypted password seems to work for postgres, but does not work for SqlServer.  I have yet to try it with Oracle.  I begin to wonder if the fact it worked with Postgres was an accident or some other environmental issue.

diracz:
I think what you might look into: http://stackoverflow.com/questions/12834604/using-encrypted-password-for-the-datasource-used-in-spring-applicationcontext-xm

DatabaeReader/Writer uses org.apache.commons.dbcp.BasicDataSource, which you can override to have getPassword/setPassword decrypt/encrypt password. 

Hope that helps.

tggagne:
I think it does...

So, if I created a subclass of org.apache.commons.dbcp.BasicDataSource I would have to specify its name in the database-conf.xml's bean for my DB connection?

If the subclass was named ObscuredPasswordDataSource the bean would look like..
```
<bean id="mypostgres" class="org.apache.commons.dbcp.ObscuredPasswordDataSource" .. >
    ...
    <property name="password" value="some encrypted value" />
</bean>
```

tggagne:
Not being a Maven expert... how can I rebuild dataloader from the command line and skip the step where it tries to "install" it on my mac?

tggagne:
diracz, I followed the advice from that stackoverflow link and created a new class, DataSource, that does the simple base64 decoding of the password.

While this obscures the password it doesn't obscure as well as dataloader's encryption/decryption of the sfdc.password.

I would prefer to use that approach, but hacking the Config class doesn't quite seem the right approach.

Do you know how I might get some help or clues if there's a better way to use Config's encryption?

The code has been committed to my repository here.

https://github.com/tggagne/dataloader/commit/e0562ab8cec8e8fb27b70452d0dcc38cfcf05cbd

tggagne:
I hacked in the UncryptUtil stuff, so the password may be decrypted.  To be /totally/ compatible with Config's use of encryption, it needs to use the keyfile -- which it doesn't yet.

[...]

PR #71, opened 2015-04-17
New BasicDataSource subclass uses EncryptionUtil to decrypt database passwords

tggagne:
For Issue #64, To use encrypted passwords (encrypt -e password) as database passwords inside database-conf.xml, all users need to do is use the classname "com.salesforce.dataloader.dao.database.DataSource" instead of "org.apache.commons.dbcp.BasicDataSource" in their DB connection bean.

An example would be:
```
<bean id="qa" class="com.salesforce.dataloader.dao.database.DataSource" destroy-method="close">
    <property name="driverClassName" value="oracle.jdbc.OracleDriver"/>
    <property name="url" value="jdbc:oracle:thin:username/@(DESCRIPTION=(ADDRESS=(PROTOCOL=TCP)(HOST=host.domain.lan)(PORT = 1521)) (CONNECT_DATA = (SERVER = DEDICATED) (SERVICE_NAME = CRXXXXA)))"/>
    <property name="username" value="username"/>
    <property name="password" value="51df4a60230cb4fc5e7cd752e605d44108e77ee68be2d911"/>
</bean>
```
This reduces the risk of sharing configuration files with other developers, and especially through repositories like "git" where everything is remembered.

### DIFF
--- a/src/main/java/com/salesforce/dataloader/dao/DataSource.java
+++ b/src/main/java/com/salesforce/dataloader/dao/DataSource.java
@@ -1,8 +1,11 @@
 package com.salesforce.dataloader.dao.database;
 
 import java.io.IOException;
-import sun.misc.BASE64Decoder;
 import org.apache.log4j.Logger;
+
+import com.salesforce.dataloader.security.EncryptionUtil;
+import com.salesforce.dataloader.config.Messages;
+import java.security.GeneralSecurityException;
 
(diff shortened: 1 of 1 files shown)


diracz:
Couple comments:
- More in-line documentation on how to use this would be better. 
- "DataSource" might not be the best class name. "EncryptedDataSource" might be better. 
Nothing major, so I'll merge this right for you to use. Please consider doing some enhancement later.

PR #72, opened 2015-04-18
Issue64 - Renamed class and included some inline doc on how to use it.

tggagne:
Ask and you shall receive... eventually.


diracz:
Nice addition. Like it!

Issue #76, opened 2015-04-28
Field mapping is invalid: OPRID => Alias, PS_OPRID__c


tggagne:
Inside LoadMapper, it seems to think the sfdc field name is "Alias, PS_OPRID__c," instead of there being two separate fields, Alias and PS_OPRID__c, each mapped to the same source field, OPRID.

Curiously, this works for literal mappings like...

```
"en_US"=LocaleSidKey, LanguageLocaleKey
```

But not

```
OPRID=Alias, PS_OPRID__c
```

tggagne:
Why should it matter whether the mapping is a constant or not?

(from Mapper.java)
```
    @Override
    protected void putPropertyEntry(Entry<Object, Object> entry) {
        String dao = (String)entry.getKey();
        String sfdc = (String)entry.getValue();
        if (isConstant(dao))
            putConstant(sfdc, dao);
        else if (!hasDaoColumns() || hasDaoColumn(dao)) putMapping(dao, sfdc);
    }
```
The problem could be that the map is built from with the source value as the key and the destination value as the value, which requires the source values be unique, or at least implies it.

Better if the destination values (sfdc fields) were the key because it would be incorrect for a single field to be the destination of two sources.  It is more likely (and useful) if a single source column can be mapped to multiple destinations fields.

tggagne:
So LoadMapper simply creates a map, leaving (in my case) oprid => "alias, operid__c."  

To get past the AbstractLoadAction.verifyMappingAreValid() we'll need that function to parse the value looking for multiple destinations, then we'll need whatever function that actually maps the data from the source to the destination to set the value of each of the destination fields.

Does that make sense to anyone else?  Does it seem a reasonable plan?

tggagne:
PartnerClient.getField(String apiName) get's a partner field, which is not a string, and sets it as the value of "oprid,", which means the map's value may have to be a list of com.sforce.soap.partner.Field.  I wonder how obnoxious that may be to change...

Does anyone know where the data mappings actually take place? I mean, when the source column's value is copied to the destination field before insert/upsert?

tggagne:
It looks like LoadMapper.mapData() could be modified to handle a list<Field> so that when it creates the new Row() it could map multiple values.  Perhaps LoadMapper.getMapping() should return the list<Field>.  It should be coded to always return a list, empty or not.  That way callers could specifically look for empty lists to detect mapping errors, or iterate through the list assigning the values.

I think I may try that.

[...]

PR #77, opened 2015-04-28
Proposed changes for issue #76

tggagne:
I should have created a separate issue for renaming the package, but it is what it is.

Lastly, I have some proposed changes so we can now map a single source column to multiple destination columns.  I've only tested it with DAOs.

So the mapping file now supports,

```
sourceField=destField1, destField2
```


src/main/java/com/salesforce/dataloader/mapping/LoadMapper.java
```diff
@@ -88,8 +92,8 @@ public void verifyMappingsAreValid() throws MappingInitializationException {
         for (Map.Entry<String, String> entry : getMappingWithUnmappedColumns(false).entrySet()) {
             String sfdcName = entry.getValue();
             if(StringUtils.hasText(sfdcName)) {
-                final Field f = getClient().getField(sfdcName);
-                if (f == null)
+                final HashSet<Field> fieldSet = getClient().getField(sfdcName);
+                if (fieldSet == null | fieldSet.isEmpty())
```
diracz: Is that a single "|" ?

tggagne: It is!

But don't let that distract you.  My changes were based off the 33.0 
branch 'cause that's the only one I can test with.  I'm unsure how to 
test v34 against my orgs.

I can fix that, but my bigger question is, though it works for what I 
need, are the maintainers OK with its implementation?  I figured 
creating a pull request was the easiest way to share the code.

rjmazzeo: Let me ask a few people about this pull request.

In the meantime, I recommend changing the HashSet to a LinkedHashSet and returning it as a Set. That way we maintain the insert order but don't let downstream users know it. We had issues with JDK8 changing the natural return order of HashSets.

diracz: If your org doesn't have v34 yet, I don't think you can test against it, but there shouldn't be anything foundationally different.
In concept I personally don't have problem with your change. It's a matter of going more thorough testing. 
In this line, can you explain why "|" instead of "||" ? "|" does not escape the second part if the first part is true, so it might create null pointer exception. Do I get it wrong?

neetikumar: Would you be able to write some tests for this

tggagne: What order does the HashSet make?  Why would they need to be in any 
order as long as they're only in the list once?  Using a HashSet to 
communicate there's no order keeps programmers from depending on 
it--which you're right isn't guaranteed.

tggagne: The | was a typo.  I'll change that.  In the examples I tested it with 
it apparently didn't cause a problem.

tggagne: That's a good question.  I don't think I've been able to get the test to 
run.  When I make it locally I have to -DskipTests. Honestly, I'm more 
of a Smalltalker and Apex programmer than Java. I needed help just to 
get the Maven project into Eclipse!

I'll try again to see if they'll run.

tggagne: So with a fresh pull from forcedotcom/dataloader, and building it with

mvn clean package -license.skip=true

I get "Tests run: 285, Failures: 210, Errors: 9, Skipped: 0"

I must be missing something in my configuration...

rjmazzeo:
What's the state of this pull request?  Did you re-submit?

tggagne:
I didn't submit another one.  I asked for help writing the unit tests.  
I'm not a regular Java developer and am unfamiliar with how Dataloader 
unit testing is set-up for mock loads.  Any help would be appreciated.

rjmazzeo:
I recall your trouble earlier. Don't worry about the unit tests at this point, we can look at them. However, can you send this pull request to the develop branch instead of main? We prefer to keep main ready for hot fixes and this doesn't quite qualify.

ashitsalesforce:
Closing the PR as it is 5+ years old and has a large number of files changed.

PR #78, opened 2015-04-29
issue #76, fixed package name and inline doc to reflect directory struct...

tggagne:
This pull request just fixes the package naming of the EncryptingDataSource class.


diracz:
I'm merging this DB fix itself first.

PR #86, opened 2015-09-28
added spring-expression as depdendency, added maven shade support.

dsipasseuth:
Hi, 

This pull request is to add spring expression language support for a more easy parameter configuration in files like process-conf.xml and database-conf.xml.

Also added Shade support so Spring dependencies in uber jar don't overlap and append nicely.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -127,6 +127,11 @@
     </dependency>
     <dependency>
       <groupId>org.springframework</groupId>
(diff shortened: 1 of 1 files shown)


rjmazzeo:
I'll do a quick build verification today and merge.

rjmazzeo:
Looks good. There is a shade conflict with the develop branch but I'll take that one on.
Tested on Windows/Mac

PR #92, opened 2015-11-09
Configurable exit code mapper for cli

dsipasseuth:
This is my contribution for adding exit code mapping to SalesForce Data Loader CLI.

Via Spring xml file, it is now possible to attach a exit code mapper. 
By default it have one with old behavior (always return 0 and return -1 on any technical error)

Happy Review


rjmazzeo:
Nice work. I just got back from being out of the office for a few weeks. So, pardon me if I take a day or two to review before merging. Thanks for adding the unit tests.

dsipasseuth:
Hi, 

Just asking for some news.

Any troubles with merging ?

rjmazzeo:
Hah, last week the madness continued on a different project. I should be able to look at this tomorrow. I apologize for the long delay on getting to this.

rjmazzeo:
Ok... from a review standpoint the only issue I want to be sure of is we we not requiring people to update their configs, or did you encounter issues with the old format?

Will pull local and review usages.

dsipasseuth:
In the bean definition, you just have to add

```xml
<bean id="someid" class="com.salesforce.dataloader.process.ProcessRunner" scope="prototype" lazy-init="true">
[....]
<property name="exitCodeMapper">
        <bean class="com.salesforce.dataloader.process.ExitCodeMapper">
        <property name="onSuccess" value="0" />
        <property name="onException" value="1" />
        <property name="onNoDataFound" value="2" />
        <property name="onNoneSucceed" value="3" />
        <property name="onPartialSuccess" value="4" />
        <property name="apiFaultExitCodeMapper">
            <map key-type="java.lang.String" value-type="java.lang.Integer">
                <entry key="INVALID_LOGIN" value="200" />
            </map>
        </property>
        <property name="asyncApiExitCodeMapper">
            <map key-type="java.lang.String" value-type="java.lang.Integer">
                <entry key="Timeout" value="300" />
            </map>
        </property>
    </bean>
    </property>
    <property name="progressMonitor">
        <bean class="com.salesforce.dataloader.action.progress.ReportCountProgressAdapter" />
    </property>
[....]
</bean>
```

rjmazzeo:
Ok... sorry about the shenanigans. I ran the integration tests against this and I got 13 errors I didn't expect. They all look fixable (maybe even trivial) but we need to fix them before merging. 
Attached is the html files that intellij creates as part of a test run. One file has the 4 expected errors and the other has the 13 additional ones.

If you don't have time then I'll make time. Let me know if you rather I do and I'll add it to my worklist.

[testresults.zip](https://github.com/forcedotcom/dataloader/files/66136/testresults.zip)

dsipasseuth:
Hi,

Sorry for the long response. didn't had a lot of time to work on testing but i've made few changes on the branch.

rjmazzeo:
Integration testing is a pain right now but doable. In the back of my mind, I want to change it to use an installable package to cut down on the 30 minutes of setup an org requires. I don't have a priority or timeline for this though. Still the directions for integration testing are on the wiki. I have the org setup on my personal box so I can resolve these issues, if you would prefer. I'm just not sure on the timing.

dsipasseuth:
The thing is, i'm having more than a hundred integration test failures (some very strange like complaining oracle_id__c is not a number...). I'm working on EU (FR) instances and don't have access to v36 api right now... i've made few changes of my own to go back to v35 (not commited)

rjmazzeo:
"oracle_id__c" is not a number sounds like your intergration setup isn't working. That's a custom field that needs to be added to your test org from the setup directions here:
https://github.com/forcedotcom/dataloader/wiki/Testing-Dataloader

I really need to make the integration testing a public metadata package.

dsipasseuth:
hum... i've left this field as a number... i've changed it to text as documentation. I should have integration test up and running soon.

zefreak:
This would be very useful, what is the status?

ashitsalesforce:
Closing the PR due to long elapsed time since the request coupled with a large number of files as part of the PR (136) and conflicts.

PR #95, opened 2015-12-15
better null handling in csv output. now a ,, instead of a ,"",

dokeeffe:
Hi,
I made a change to the CSV writer if you would like to take it. It writes ,, instead of ,"", for nulls.
This will help when importing the CSV into a relational database as the "" causes problems with non char/varchar columns in the target table.
Derek


rjmazzeo:
Looks good. I'll need to run in through the integration test suite. I'll do that and let you know.

ashitsalesforce:
This is a major change touching 300+ files and that too requested 5+ years ago. Closing this PR because it is not possible to review in its current form.

PR #102, opened 2016-02-02
encrypt and process batches needed updated paths for java

rjmazzeo:
we decoupled java but the paths in the batches were still expecting java to be included. Updated.

### DIFF
--- a/src/main/nsis/bin/process.bat
+++ b/src/main/nsis/bin/process.bat
@@ -24,6 +24,25 @@ goto end
 set PROCESS_OPTION=
 if not [%2]==[] set PROCESS_OPTION=process.name=%2
 
-..\Java\bin\java.exe -cp ..\${pom.build.finalName}-uber.jar -Dsalesforce.config.dir=%1 com.salesforce.dataloader.process.ProcessRunner %PROCESS_OPTION%
+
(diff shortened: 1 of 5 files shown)


src/main/nsis/bin/process.bat
```diff
@@ -24,6 +24,25 @@ goto end
...
+        IF EXIST "%%i" (
+            set JAVA_HOME=%%i
+        ) ELSE (
+            echo %%i
+        )
+    )
+)
+
+IF "%JAVA_HOME%" == "" (
+    echo Please set JAVA_HOME before running the encrypt.bat
```
apoorvashenoy: I get that this is a added layer of notification, but isn't this auto-generated?

rjmazzeo: No.. line 33 will say something like "This application requires JAVA 1.8 32bit" but it won't say anything about JAVA_HOME. 

We can omit it since if the user doesn't set the system will try to figure it out. I'll ask Tess if it's noise for users.

apoorvashenoy: If anything, it seems to add more help. 
I'll approve the merge request and you can make the edits based on Tess's feedback.

PR #103, opened 2016-02-03
Oauth ui fixes

rjmazzeo:
Fix ui labels and add support for seperate oauth clientids for bulk and partner connections


src/main/java/com/salesforce/dataloader/ui/Grid12.java
```diff
@@ -165,4 +172,11 @@ public void show(Control control) {
     public void pack() {
         composite.pack();
     }
+
+    private void applyFormatting(Control control) {
+        //use this to apply standard formatting
+
```
apoorvashenoy: Do you mean to have this comments only?

rjmazzeo: yeah. I put all this code here for debugging. We don't need any standard formatting but I don't want to create it again next time I need to see the grids on the UI. I can pull it if you think it's better.

apoorvashenoy: Makes sense. Let's leave it in here.

PR #107, opened 2016-07-19
use ApplicationContext instead of XmlBeanFactory

r-nasiri:
This change will allow using variables in the SQL's that are defined in database config file. The variables can be set in config.properties or as an environment variable.

### DIFF
--- a/src/main/java/com/salesforce/dataloader/dao/database/DatabaseConfig.java
+++ b/src/main/java/com/salesforce/dataloader/dao/database/DatabaseConfig.java
@@ -26,8 +26,8 @@
 package com.salesforce.dataloader.dao.database;
 
 import org.apache.commons.dbcp.BasicDataSource;
-import org.springframework.beans.factory.xml.XmlBeanFactory;
-import org.springframework.core.io.FileSystemResource;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.support.FileSystemXmlApplicationContext;
 
 public class DatabaseConfig {
     private BasicDataSource dataSource;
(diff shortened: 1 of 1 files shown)


rjmazzeo:
Sorry...we've been a bit busy. Should have bandwidth next week to look at this. I'll run in through the existing integration tests and let you know.

src/main/java/com/salesforce/dataloader/dao/database/DatabaseConfig.java
```diff
@@ -26,8 +26,8 @@
 package com.salesforce.dataloader.dao.database;
 
 import org.apache.commons.dbcp.BasicDataSource;
-import org.springframework.beans.factory.xml.XmlBeanFactory;
-import org.springframework.core.io.FileSystemResource;
+import org.springframework.context.ApplicationContext;
```
bigassforce: Enables `PropertyPlaceholderConfigurer` to be used in **database-conf.xml**

Then the data source can load variables like `${db.username}` from a config file?

```
<!DOCTYPE beans PUBLIC "-//SPRING//DTD BEAN//EN" "http://www.springframework.org/dtd/spring-beans.dtd">
<beans>
    <bean id="propertyPlaceholderConfigurer" class="org.springframework.beans.factory.config.PropertyPlaceholderConfigurer">
        <property name="locations">
            <list>
                <value>classpath:database.properties</value>
            </list>
        </property>
    </bean>
    <bean class="org.apache.commons.dbcp.BasicDataSource" destroy-method="close" id="MsqlDataSource">
        <property name="username" value="${db.user}"/>
        ...
```

http://forum.spring.io/forum/spring-projects/container/58380-propertyplaceholderconfigurer-is-ignored

ashitsalesforce:
Ran relevant integration tests (com.salesforce.dataloader.dao.database.DatabaseTest and com.salesforce.dataloader.process.DatabaseProcessTest) successfully on Windows and Mac. Another reason to make this code change is that XmlBeanFactory is deprecated (https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/beans/factory/xml/XmlBeanFactory.html).

PR #109, opened 2016-08-08
Updated Bulk API error handling

rjmazzeo:
Revised bulk load to not catastrophically fail if batch contains errors. Also changed abstract action exit code to make sure it does it's best to report success or failures

### DIFF
--- a/src/main/java/com/salesforce/dataloader/action/AbstractAction.java
+++ b/src/main/java/com/salesforce/dataloader/action/AbstractAction.java
@@ -26,6 +26,7 @@
 
 package com.salesforce.dataloader.action;
 
+import java.util.ArrayList;
 import java.util.LinkedList;
 import java.util.List;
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/salesforce/dataloader/action/AbstractAction.java
```diff
@@ -128,24 +130,40 @@ public final void execute() {
...
-                getMonitor().doneSuccess(getMessage("success", args)); //$NON-NLS-1$
-            }
 
         } catch (final Exception e) {
-            handleException(e);
+            exceptions.add(e);
         } finally {
-            closeAll();
+            try{
+                flush(); //make sure we don't early abort here
```
jjangsfdc: What was the intention of breaking flush() and closeAll() into two try catch?

rjmazzeo: closeAll doesn't flush the buffers. If the visitors threw there could be unwritten lines in the buffer.

Other option is to change closeAll to flush.

jjangsfdc: Oh I was mistaken for seeing flush() as outside of the finally block. I see what you are doing.

PR #114, opened 2016-08-18
Connection clean up and upgrade httpClient

jjangsfdc:
Upgrade Apache httpClient lib from 4.2.1 to 4.5.2 and update depreciated API calls with connection cleanups.
W-3238933

### DIFF
--- a/src/main/java/com/salesforce/dataloader/client/DefaultSimplePost.java
+++ b/src/main/java/com/salesforce/dataloader/client/DefaultSimplePost.java
@@ -31,16 +31,17 @@ import com.salesforce.dataloader.exception.ParameterLoadException;
 import com.sforce.ws.tools.VersionInfo;
 import org.apache.commons.io.IOUtils;
 import org.apache.http.HttpHost;
-import org.apache.http.HttpResponse;
 import org.apache.http.auth.AuthScope;
 import org.apache.http.auth.Credentials;
 import org.apache.http.auth.NTCredentials;
 import org.apache.http.auth.UsernamePasswordCredentials;
+import org.apache.http.client.CredentialsProvider;
+import org.apache.http.client.config.RequestConfig;
 import org.apache.http.client.entity.UrlEncodedFormEntity;
+import org.apache.http.client.methods.CloseableHttpResponse;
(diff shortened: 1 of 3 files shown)


src/main/java/com/salesforce/dataloader/client/DefaultSimplePost.java
```diff
@@ -72,62 +73,80 @@
...
-            client.getCredentialsProvider().setCredentials(scope, credentials);
-        }
-
+        CloseableHttpClient client = HttpClients.createDefault();
         try {
-            if (ntlmDomain.length() > 0) {
-                // need to send a HEAD request to trigger NTLM authentication
-                HttpHead head = new HttpHead("http://salesforce.com");
-                client.execute(head);
-                head.releaseConnection();
```
rjmazzeo: We don't need to issue a head request for nt proxy servers anymore?

rjmazzeo: never mind... I seeit.

src/main/java/com/salesforce/dataloader/client/HttpClientTransport.java
```diff
@@ -90,64 +95,83 @@ public OutputStream connect(String url, String soapAction) throws IOException {
...
+                }
+
+            } finally {
+                if (response != null)
+                    response.close();
             }
-   
         } finally {
-            post.releaseConnection();
+            client.close();
```
xbiansf: Looks good.  Like the new way of closing connection.

jjangsfdc: :) Sorry the diffs are kind of hard to follow. Adding Try block messed around it a lot!

rjmazzeo:
I think it looks good enough to try. just see what happens with intergration tests.

We should also need to vet the  proxy changes

jjangsfdc:
I've already ran the integration tests and all passed.

apoorvashenoy:
I would test OAuth as well. But Rocky's tests should cover those.

apoorvashenoy:
Not that I am aware of.

PR #115, opened 2016-08-22
Improve httpclient with SNI support

xbiansf:
Improve http client from potential client leak and adding verification test for Server name indicator in SSL.

### DIFF
--- a/src/main/java/com/salesforce/dataloader/client/DefaultSimplePost.java
+++ b/src/main/java/com/salesforce/dataloader/client/DefaultSimplePost.java
@@ -29,6 +29,7 @@ package com.salesforce.dataloader.client;
 import com.salesforce.dataloader.config.Config;
 import com.salesforce.dataloader.exception.ParameterLoadException;
 import com.sforce.ws.tools.VersionInfo;
+
 import org.apache.commons.io.IOUtils;
 import org.apache.http.HttpHost;
 import org.apache.http.auth.AuthScope;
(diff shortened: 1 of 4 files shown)


rjmazzeo:
Nice job. This coupled with Jae's edit should really solidy http handling.

src/main/java/com/salesforce/dataloader/client/DefaultSimplePost.java
```diff
@@ -73,80 +74,68 @@
...
-                        response = client.execute(head);
-                    } finally {
-                        response.close();
-                    }
+            CredentialsProvider credentialsprovider = new BasicCredentialsProvider();
+            credentialsprovider.setCredentials(scope, credentials);
+            httpClientBuilder.setDefaultCredentialsProvider(credentialsprovider).build();
+        }
+
+        try (CloseableHttpClient httpClient = httpClientBuilder.build()) {
```
jjangsfdc: Does it dispose/close itself after the Try block?

xbiansf: Yes.  All the try resource block will call close whenever there is any exception.  If there is any exception inside close(), that exception will be suppressed which is what we want most of the time. This also releases the resource from previous HttpClients.createDefault() as well by using the builder pattern.

jjangsfdc: Nice. Please make sure to run the integration tests before we pull in the change.

PR #116, opened 2016-08-25
Support more control on csv separator characters

xbiansf:
1. We will support default comma and Tab
2. Give user to select/deselect comma, Tab and other character.
3. Test added


xbiansf:
Changed the UI text according to feedback.

PR #119, opened 2016-09-16
Use CSV reader with different setting for system csv from customer csv

xbiansf:


xbiansf:
Already passed integration test.

PR #125, opened 2016-12-07
Fix url redirected after cancelation

xbiansf:
This is fix a customer case, where user logged in using Oauth, but switched the user name and cancel it,  then the default host address is changed to the blitz url.

### DIFF
--- a/src/main/java/com/salesforce/dataloader/model/LoginCriteria.java
+++ b/src/main/java/com/salesforce/dataloader/model/LoginCriteria.java
@@ -96,7 +96,6 @@ public class LoginCriteria {
     public void updateConfig(Config config) {
         config.setValue(Config.USERNAME, config.STRING_DEFAULT);
         config.setValue(Config.PASSWORD, config.STRING_DEFAULT);
-        config.setValue(Config.ENDPOINT, config.STRING_DEFAULT);
         config.setValue(Config.SFDC_INTERNAL_IS_SESSION_ID_LOGIN, false);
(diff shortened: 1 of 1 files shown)


src/main/java/com/salesforce/dataloader/model/LoginCriteria.java
```diff
@@ -96,7 +96,6 @@ public void setSessionId(String sessionId) {
     public void updateConfig(Config config) {
         config.setValue(Config.USERNAME, config.STRING_DEFAULT);
         config.setValue(Config.PASSWORD, config.STRING_DEFAULT);
-        config.setValue(Config.ENDPOINT, config.STRING_DEFAULT);
```
jjangsfdc: How does this fix the issue?

xbiansf:
Jae, see my overall comment above. It is a little bit complex.  More specific details:  with setting the endpoint to empty string(STRING_DEFAULT),  it will update the config.  Then the next time Setting dialog or log in dialog with local a null string, then it will use a default string from partner API.  With the fix, will keep the last good endpoint, so you can still login correctly.    :-)

src/main/java/com/salesforce/dataloader/model/LoginCriteria.java
xbiansf: Jae, see my overall comment above. It is a little bit complex. More specific details: with setting the endpoint to empty string(STRING_DEFAULT), it will update the config. Then the next time Setting dialog or log in dialog with local a null string, then it will use a default string from partner API. With the fix, will keep the last good endpoint, so you can still login correctly. :-)

PR #129, opened 2017-02-27
Fix an issue when CLI will fail to find extracted columns while exporting to DB

xbiansf:
Fixing SF ticket W-3732361.   Dataloader command line exports objects in  windows command line environment, it will complain columns names not found while uploading to external DB.

### DIFF
--- a/src/main/java/com/salesforce/dataloader/model/Row.java
+++ b/src/main/java/com/salesforce/dataloader/model/Row.java
@@ -30,6 +30,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


src/main/java/com/salesforce/dataloader/model/Row.java
```diff
@@ -50,7 +51,7 @@ public Row() {
     }
 
     public Row(int columnCount) {
-        internalMap = new HashMap<String, Object>(columnCount);
+        internalMap = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
```
jjangsfdc: Good to know. Thanks

PR #131, opened 2017-04-12
Adding default value for Integer for importing

xbiansf:


src/main/java/com/salesforce/dataloader/dyna/SforceDynaBean.java
```diff
@@ -313,7 +313,7 @@ synchronized static public void registerConverters(Config cfg) {
         // Register DynaBean type conversions
         ConvertUtils.register(new DateConverter(tz, useEuroDates), Calendar.class);
         ConvertUtils.register(new DoubleConverter(), Double.class);
-        ConvertUtils.register(new IntegerConverter(), Integer.class);
+        ConvertUtils.register(new IntegerConverter(null), Integer.class);
```
xbiansf: This is used to process "#N/A" when mapped integer string to an Integer Data type in salesforce.  Given this change,  the IntegerConverter will have a default value,  and use null when it find #N/A in IntergerConverter.java#convert() method.

jjangsfdc: Do we want to set default as null or 0? Maybe I am not fully understanding the problem here...

xbiansf: null means no value provided.  0 means 0. So null should be the correct value.

jjangsfdc: I see, so this is the case when nothing is set.

PR #132, opened 2017-05-15
Upgrade to V40

xbiansf:
We are upgrading to V40


apoorvashenoy:
Looks good to me. Will merge.

PR #137, opened 2017-06-21
ProcessRunner: set the last run date to the time the process was started

floppym:
This prevents the last run date from being skewed due to long-running
processes.


jjangsfdc:
Merging https://github.com/forcedotcom/dataloader/pull/137

PR #138, opened 2017-06-23
Add support for PK Chunking of bulk queries

floppym:


ashitsalesforce:
Hi @floppym, I am closing this PR as it has conflicts with the code. Most of the proposed changes are incorporated that should enable a relatively smaller code change to support PK chunking for bulk query when we decide to support this feature enhancement.

PR #142, opened 2017-07-20
added config for DuplicateRuleHeader

tee-tee-dev:
Sometimes we want to insert records, without duplicate rules errors.
we added config options for DuplicateRuleHeader.


ashitsalesforce:
Thanks for the contribution. Will merge it.

PR #145, opened 2017-09-19
add Linux command line interface for batch processing

danchal:
Please find a Linux equivalent for encrypt.bat and process.bat for your review.
Cheers,
Dan

### DIFF
--- a/src/main/nsis/bin/process.sh
+++ b/src/main/nsis/bin/process.sh
@@ -1,4 +1,4 @@
-#/bin/bash
+#!/bin/bash
 
 function error {
     echo "$1" >&2
@@ -25,6 +25,9 @@ function usage {
     exit 1
 }
 
+# set execution path
+EXE_PATH=$( dirname "${BASH_SOURCE[0]}" )
+
 if [ "$#" -eq 1 ]; then
     PROCESS_OPTION=""
 elif [ "$#" -eq 2 ]; then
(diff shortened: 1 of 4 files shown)


src/main/nsis/bin/process.sh
```diff
@@ -0,0 +1,47 @@
+#/bin/bash
+# change to directory of this script
+cd "$( dirname "${BASH_SOURCE[0]}" )"
```
cbhacking: The Windows scripts don't change the working directory. I can see why you're doing it - so the relative path to the .JAR is correct - but that could be better handled with a local variable.

danchal: Thanks for the review, you are correct, the behaviour must be the same as the Windows scripts. I will modify the PR.

danchal:
I updated the [PR](https://github.com/forcedotcom/dataloader/pull/145/commits/47a327d63cc27d57381718f01c6c2a84f3cd9470), removing the change program directory lines. The behaviour is now the same as the Windows scripts.

Reading your last comment, are you suggesting that both the Windows and Linux scripts be now modified to handle the .JAR path better? I think that this would be an improvement on the current behaviour, if implemented as you have suggested.

ashitsalesforce:
While we do not support linux CLI, you may be able to do so in v52 onwards (or try it in v51 if you compile from the source) by using the following options:

`java -jar dataloader-51.0.1-uber.jar run.mode=batch process.name=<process name> salesforce.config.dir=<path to config dir containing process-conf.xml and config.properties files>`

PR #147, opened 2017-12-15
Xbian/fix mapping issue

xbiansf:
1. Upgrade to V42.
2.  Fixing the mapping issue where if you have a csv aggregated field name ending with  another field name in the mapping page.

### DIFF
--- a/src/main/java/com/salesforce/dataloader/mapping/Mapper.java
+++ b/src/main/java/com/salesforce/dataloader/mapping/Mapper.java
@@ -174,18 +174,23 @@ public abstract class Mapper {
     }
 
     public String getMapping(String srcName) {
-        if(map.containsKey(srcName)) {
+        return getMapping(srcName, false);
+    }
+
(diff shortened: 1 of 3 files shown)


src/main/java/com/salesforce/dataloader/mapping/Mapper.java
```diff
@@ -173,19 +173,28 @@ protected boolean hasMappings() {
...
+
+    public String getMapping(String srcName, boolean strictMatching) {
+        if (map.containsKey(srcName)) {
             return map.get(srcName);
         }
         // handle aggregate queries
-        for(Entry<String, String> entry: map.entrySet()) {
-            if(entry.getKey().endsWith("." + srcName)) {
-                return entry.getValue();
+        if (!strictMatching) {
```
xbiansf: using strictMatching to limit the change only to mapping page scenario.   The root cause of the bug is because one field name, say "xyz"  and another field "abc.xyz", during refresh, it will think xyz is already mapped, when it see *.xyz.

PR #149, opened 2018-01-03
Fixing Mapping issue part 2

xbiansf:
Fixing the partner API call mapping issue where fuzzy mapping will cause "name" to "company  name" mapping.

### DIFF
--- a/src/main/java/com/salesforce/dataloader/mapping/LoadMapper.java
+++ b/src/main/java/com/salesforce/dataloader/mapping/LoadMapper.java
@@ -26,15 +26,18 @@
 
 package com.salesforce.dataloader.mapping;
 
-import java.util.*;
-import java.util.Map.Entry;
-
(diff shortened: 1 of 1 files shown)


src/main/java/com/salesforce/dataloader/mapping/LoadMapper.java
```diff
@@ -73,7 +76,7 @@ protected void putPropertyEntry(Entry<Object, Object> entry) {
     public Row mapData(Row localRow) {
         Row mappedData = new Row();
         for (Map.Entry<String, Object> entry : localRow.entrySet()) {
-            String sfdcName = getMapping(entry.getKey());
+            String sfdcName = getMapping(entry.getKey(), true);
```
xbiansf: This was used by partner api to map from csv column name to sdfc name, we need to exact mapping to process this as well.

PR #150, opened 2018-02-07
Xbian/security patch

xbiansf:
Adding AES encryption 
Fixing the date bug
Nullify the password saved in config.properties.

### DIFF
--- a/src/main/java/com/salesforce/dataloader/config/Config.java
+++ b/src/main/java/com/salesforce/dataloader/config/Config.java
@@ -26,6 +26,13 @@
 
 package com.salesforce.dataloader.config;
 
+import com.salesforce.dataloader.action.OperationInfo;
+import com.salesforce.dataloader.exception.ConfigInitializationException;
+import com.salesforce.dataloader.exception.ParameterLoadException;
+import com.salesforce.dataloader.security.EncryptionAesUtil;
+
+import org.apache.log4j.Logger;
+
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileOutputStream;
@@ -34,19 +41,24 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.io.PrintStream;
 import java.io.PrintWriter;
+import java.io.UnsupportedEncodingException;
 import java.nio.charset.Charset;
 import java.security.GeneralSecurityException;
 import java.text.DateFormat;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
-import java.util.*;
-
-import org.apache.log4j.Logger;
-
-import com.salesforce.dataloader.action.OperationInfo;
-import com.salesforce.dataloader.exception.ConfigInitializationException;
-import com.salesforce.dataloader.exception.ParameterLoadException;
-import com.salesforce.dataloader.security.EncryptionUtil;
+import java.util.ArrayList;
+import java.util.Calendar;
+import java.util.Collections;
+import java.util.Date;
+import java.util.Enumeration;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Properties;
+import java.util.StringJoiner;
+import java.util.TimeZone;
 
 /**
  * @author Lexi Viripaeff
@@ -229,7 +241,7 @@ public class Config {
      * <code>encrypter</code> is a utility used internally in the config for reading/writing
      * encrypted values. Right now, the list of encrypted values is known to this class only.
      */
-    private final EncryptionUtil encrypter = new EncryptionUtil();
+    private final EncryptionAesUtil encrypter = new EncryptionAesUtil();
 
     private boolean isBatchMode = false;
 
@@ -651,15 +663,20 @@ public class Config {
         initEncryption(propMap);
 
         // decrypt encrypted values
-        if (propMap.containsKey(PASSWORD)) {
-            String propValue = decryptProperty(encrypter, propMap, PASSWORD, isBatchMode());
+        decryptPasswordProperty(values, PASSWORD);
+        decryptPasswordProperty(values, PROXY_PASSWORD);
+        decryptPasswordProperty(values, OAUTH_ACCESSTOKEN);
+        decryptPasswordProperty(values, OAUTH_REFRESHTOKEN);
+
+    }
+
+    private void decryptPasswordProperty(Map values, String propertyName) throws ConfigInitializationException {
+        Map<String, String> propMap = values;
+        // initialize encryption
+        if (propMap.containsKey(propertyName)) {
+            String propValue = decryptProperty(encrypter, propMap, propertyName, isBatchMode());
             if (propValue == null) propValue = "";
-            propMap.put(PASSWORD, propValue);
-        }
-        if (propMap.containsKey(PROXY_PASSWORD)) {
-            String propValue = decryptProperty(encrypter, propMap, PROXY_PASSWORD, isBatchMode());
-            if (propValue == null) propValue = "";
-            propMap.put(PROXY_PASSWORD, propValue);
+            propMap.put(propertyName, propValue);
         }
     }
 
@@ -690,12 +707,12 @@ public class Config {
      *
      * @return decrypted property value
      */
-    static private String decryptProperty(EncryptionUtil encrypter, Map<String, String> propMap, String propName, boolean isBatch)
+    static private String decryptProperty(EncryptionAesUtil encrypter, Map<String, String> propMap, String propName, boolean isBatch)
             throws ParameterLoadException {
         String propValue = propMap.get(propName);
         if (propValue != null && propValue.length() > 0) {
             try {
-                return encrypter.decryptString(propValue);
+                return encrypter.decryptMsg(propValue);
             } catch (GeneralSecurityException e) {
                 // if running in the UI, we can ignore encryption errors
                 if (isBatch) {
@@ -725,7 +742,7 @@ public class Config {
         if (keyFile != null && keyFile.length() != 0) {
             try {
                 encrypter.setCipherKeyFromFilePath(keyFile);
-            } catch (IOException e) {
+            } catch (Exception e) {
                 String errMsg = Messages.getFormattedString("Config.errorSecurityInit", new String[]{keyFile,
                         e.getMessage()});
                 logger.error(errMsg);
@@ -793,7 +810,6 @@ public class Config {
         // no great way to do this,
         String oldPassword = encryptProperty(PASSWORD);
         String oldProxyPassword = encryptProperty(PROXY_PASSWORD);
-
         String oauthAccessToken = getString(OAUTH_ACCESSTOKEN);
         String oauthRefreshToken = getString(OAUTH_REFRESHTOKEN);
         putValue(OAUTH_ACCESSTOKEN, "");
@@ -831,10 +847,11 @@ public class Config {
      * @param propName name of the property
      * @return old value of the property
      */
-    private String encryptProperty(String propName) throws GeneralSecurityException {
+    private String encryptProperty(String propName) throws GeneralSecurityException, UnsupportedEncodingException {
         String oldValue = getString(propName);
         if (oldValue != null && oldValue.length() > 0) {
-            putValue(propName, encrypter.encryptString(oldValue));
+
+            putValue(propName, encrypter.encryptMsg(oldValue));
         }
         return oldValue;
     }
--- a/src/main/java/com/salesforce/dataloader/config/Config.java
+++ b/src/main/java/com/salesforce/dataloader/config/Config.java
@@ -749,6 +749,7 @@ public class Config {
                 throw new ConfigInitializationException(errMsg);
             }
         }
+
     }
 
     /**
@@ -812,6 +813,8 @@ public class Config {
         String oldProxyPassword = encryptProperty(PROXY_PASSWORD);
         String oauthAccessToken = getString(OAUTH_ACCESSTOKEN);
         String oauthRefreshToken = getString(OAUTH_REFRESHTOKEN);
+        putValue(PASSWORD, "");
+        putValue(PROXY_PASSWORD, "");
         putValue(OAUTH_ACCESSTOKEN, "");
         putValue(OAUTH_REFRESHTOKEN, "");
 
--- a/src/main/java/com/salesforce/dataloader/security/EncryptionAesUtil.java
+++ b/src/main/java/com/salesforce/dataloader/security/EncryptionAesUtil.java
@@ -1,0 +1,195 @@
+/*
+ * Copyright (c) 2015, salesforce.com, inc.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without modification, are permitted provided
+ * that the following conditions are met:
+ *
+ *    Redistributions of source code must retain the above copyright notice, this list of conditions and the
+ *    following disclaimer.
+ *
+ *    Redistributions in binary form must reproduce the above copyright notice, this list of conditions and
+ *    the following disclaimer in the documentation and/or other materials provided with the distribution.
+ *
+ *    Neither the name of salesforce.com, inc. nor the names of its contributors may be used to endorse or
+ *    promote products derived from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED
+ * WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A
+ * PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED
+ * TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+ * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+package com.salesforce.dataloader.security;
+
+import org.apache.log4j.Logger;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.UnsupportedEncodingException;
+import java.nio.file.Paths;
+import java.security.GeneralSecurityException;
+import java.security.InvalidAlgorithmParameterException;
+import java.security.InvalidKeyException;
+import java.security.NoSuchAlgorithmException;
+import java.security.SecureRandom;
+import java.security.spec.InvalidKeySpecException;
+import java.security.spec.InvalidParameterSpecException;
+
+import javax.crypto.BadPaddingException;
+import javax.crypto.Cipher;
+import javax.crypto.IllegalBlockSizeException;
+import javax.crypto.NoSuchPaddingException;
+import javax.crypto.spec.IvParameterSpec;
+import javax.crypto.spec.SecretKeySpec;
+
+/**
+ * com.salesforce.dataloader.security
+ *
+ * @author xbian
+ */
+public class EncryptionAesUtil {
+
+    private static final Logger LOGGER = Logger.getLogger(EncryptionAesUtil.class);
+
+    // Support single text encryption and decryption
+
+    static public Cipher cipher;
+    static public byte[] cipherKey;
+
+
+    public static final int ENCRYPTION_KEY_LENGTH_IN_BYTES = 16;
+    public static final String DEFAULT_KEYFILE_NAME = "dataLoader.key";
+
+    static {
+        try {
+            cipher = Cipher.getInstance("AES/CBC/PKCS5Padding");
+
+        } catch (Exception e) {
+            LOGGER.error("Fail to initialize encryption: " + e.getMessage());
+            throw new  RuntimeException("Fail to initialize encryption:  ", e);
+        }
+    }
+
+
+    private byte[] extractIvBytes(byte[] cipheredText) {
+        byte[] iv = new byte[ENCRYPTION_KEY_LENGTH_IN_BYTES];
+        System.arraycopy(cipheredText, 0, iv, 0, ENCRYPTION_KEY_LENGTH_IN_BYTES);
+        return iv;
+    }
+
+    private byte[] extractCipherContent(byte[] cipheredText) {
+        byte[] cipheredPassword = new byte[cipheredText.length - ENCRYPTION_KEY_LENGTH_IN_BYTES];
+        System.arraycopy(cipheredText, ENCRYPTION_KEY_LENGTH_IN_BYTES, cipheredPassword, 0, cipheredText.length - ENCRYPTION_KEY_LENGTH_IN_BYTES);
+        return cipheredPassword;
+
+    }
+
+    private byte[] concatenateByteArray(byte[] a, byte[] b) {
+
+        byte[] c = new byte[a.length + b.length];
+        System.arraycopy(a, 0, c, 0, a.length);
+        System.arraycopy(b, 0, c, a.length, b.length);
+        return c;
+    }
+
+    public byte[] generateEncryptionKey() {
+        // The secure random uses threadId and tick to make sure it is not repeatable
+        SecureRandom random = new SecureRandom();
+        byte bytes[] = new byte[ENCRYPTION_KEY_LENGTH_IN_BYTES];
+        random.nextBytes(bytes);
+        return bytes;
+    }
+
+    public String createUserProfileKeyName() {
+        String path = Paths.get(System.getProperty("user.home"), "dataloader").toString();
+        File customDir = new File(path);
+
+        if (customDir.exists()) {
+            LOGGER.info(customDir + " exists");
+        } else if (customDir.mkdirs()) {
+            LOGGER.info(customDir + " was created");
+        } else {
+            LOGGER.info(customDir + " was not created");
+            throw new RuntimeException("Cannot create directory:" + path);
+        }
+        return Paths.get(path, DEFAULT_KEYFILE_NAME).toString();
+    }
+
+    public String createKeyFile(String filePath) throws IOException {
+        if (filePath == null || filePath.isEmpty()) {
+            filePath = createUserProfileKeyName();
+        }
+        byte[] key = generateEncryptionKey();
+        try (FileOutputStream fos = new FileOutputStream(filePath)) {
+            fos.write(key);
+        }
+        return filePath;
+    }
+
+    public void resetCipherKey(){
+        cipherKey = null;
+    }
+
+    public String setCipherKeyFromFilePath(String filePath) throws IOException, GeneralSecurityException {
+        byte[] data = new byte[1024];
+        try (FileInputStream fis = new FileInputStream(filePath)) {
+            int size = fis.read(data);
+            if (size < ENCRYPTION_KEY_LENGTH_IN_BYTES)
+                throw new GeneralSecurityException("Keyfile content is too short:" + filePath);
+            else
+            {
+                cipherKey = new byte[ENCRYPTION_KEY_LENGTH_IN_BYTES];
+                System.arraycopy(data, 0, cipherKey, 0, ENCRYPTION_KEY_LENGTH_IN_BYTES);
+            }
+            return new String(cipherKey, "UTF-8");
+        }
+    }
+
+    public String encryptMsg(String msg) throws GeneralSecurityException {
+        try {
+            return EncryptionUtil.bytesToText(encryptMsg(msg, cipherKey));
+        } catch (Exception e) {
+            LOGGER.error("Fail to encrypt message: " + e.getMessage());
+            throw new  GeneralSecurityException("Error to encrypt message: ", e);
+        }
+    }
+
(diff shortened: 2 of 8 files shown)


src/main/java/com/salesforce/dataloader/security/EncryptionUtil.java
```diff
@@ -149,17 +164,17 @@ private static Cipher createCipher() throws NoSuchAlgorithmException, NoSuchPadd
...
      */
     public String encryptString(String clearText) throws GeneralSecurityException {
         if (clearText == null) return clearText;
         Key oKey = getCryptoKey();
         Cipher oCipher = createCipher();
         byte[] encryptedBytes;
 
-        if (oCipher == null) { return clearText; }
+        if (oCipher == null) {
+            return clearText;
```
cbhacking: I didn't notice this before, and this whole function needs to go away anyway, but just for the record: don't do this! The correct thing to do if you can't encrypt the string is to throw an exception. If for some reason you *must* return a value, return `null`. Returning the plaintext string as the output of an encryption function under *any* circumstance creates a risk of exposing the secret data that is supposed to be protected.

src/main/java/com/salesforce/dataloader/security/EncryptionUtil.java
```diff
@@ -183,44 +198,35 @@ public String encryptString(String clearText) throws GeneralSecurityException {
...
      *
-     * @param cipherText
      * @return The decrypted string
-     * @throws NoSuchPaddingException
-     * @throws NoSuchAlgorithmException
-     * @throws BadPaddingException
-     * @throws IllegalBlockSizeException
-     * @throws InvalidKeyException
      */
     public String decryptString(String cipherText) throws GeneralSecurityException {
```
cbhacking: Unless supporting migration of existing data, get rid of this function or replace it with a wrapper around its AES-based replacement.

src/main/java/com/salesforce/dataloader/security/EncryptionUtil.java
```diff
@@ -270,12 +276,16 @@ public void setCipherKeyFromFilePath(String keyFilename) throws IOException {
     private static void printUsage() {
         LOGGER.info("\nUtility to encrypt a string based on a static or a provided key");
         LOGGER.info("Options (mutually exclusive - use one at a time): \n"
-                + "\t-g <seed text>                                 Generate key based on seed\n"
-                + "\t-v <encrypted> <decrypted value> [Path to Key] Validate whether decryption of encrypted value matches the decrypted value, optionally provide key file\n"
-                + "\t-e <plain text> [Path to Key]                  Encrypt a plain text value, optionally provide key file (generate key using option -g)");
+                + "To be deprecated:"
+                + "\t\t <seed text>                                 Generate key based on seed\n"
+                + "\t\t<encrypted> <decrypted value> [Path to Key] Validate whether decryption of encrypted value matches the decrypted value, optionally provide key file\n"
+                + "\t\t-e <plain text> [Path to Key]                  Encrypt a plain text value, optionally provide key file (generate key using option -g\n)"
```
cbhacking: Remove all support for generating DES keys (especially since the method used isn't even a secure way to generate any kind of key), and for encrypting using DES. Either replace these functions with a migrate utility (though that should just happen automatically, instead) or make them use AES instead (and don't generate the AES keys from a password, do it using raw output of `SecureRandom` and do it automatically).

src/main/nsis/samples/conf/process-conf.xml
```diff
@@ -36,10 +36,10 @@
             <map>
                 <entry key="sfdc.endpoint" value="https://login.salesforce.com"/> 
                 <entry key="sfdc.username" value="user@mycompany.com"/>
-                <!-- password below has been encrypted using key file, therefore it will not work without the key setting: process.encryptionKeyFile
-                the password is not a valid encrypted value, please generate the real value using encrypt.bat utility -->
+                <!-- password, specified below is invalid, please generate one entry in keystore only readable by yourself using the encrypt.bat utility -->
                 <entry key="sfdc.password" value="5555555555555555"/>
-                <entry key="process.encryptionKeyFile" value="c:\dataloader\samples\conf\sample.key"/>
+                <entry key="process.encryptionKeyFile" value="c:\users\username\sample.key"/>
```
cbhacking: The keyfile should be generated automatically, at a secure default path. A comment explaining this might be good.

src/main/nsis/samples/conf/process-conf.xml
```diff
@@ -63,7 +63,7 @@
                 <entry key="sfdc.debugMessagesFile" value="c:\dataloader\samples\status\sfdcSoapTrace.log"/>
                 <entry key="sfdc.endpoint" value="https://login.salesforce.com"/>
```
cbhacking: Is it actually necessary to allow changing the login endpoint? Seems like a needlessly risky thing to allow, as it could result in somebody sending their credentials to a third party or accidentally sending them in plain-text HTTP (actually, non-HTTPS should probably cause an error).

src/main/nsis/samples/conf/process-conf.xml
```diff
@@ -88,8 +88,10 @@
...
                 <entry key="sfdc.username" value="user@mycompany.com"/>
-                <!-- password specified below is invalid, please generate one using the encrypt.bat utility -->
+                <!-- password, specified below is invalid, please generate one entry in keystore only readable by yourself using the encrypt.bat utility -->
                 <entry key="sfdc.password" value="1111111111111111"/>
+                <entry key="process.encryptionKeyFile" value="c:\users\username\sample.key"/>
```
cbhacking: Same comment regarding key file and so on as above.

src/main/nsis/samples/conf/process-conf.xml
```diff
@@ -63,7 +63,7 @@
                 <entry key="sfdc.debugMessagesFile" value="c:\dataloader\samples\status\sfdcSoapTrace.log"/>
                 <entry key="sfdc.endpoint" value="https://login.salesforce.com"/>
                 <entry key="sfdc.username" value="user@mycompany.com"/>
-                <!-- password specified below is invalid, please generate one using the encrypt.bat utility -->
+                <!-- password, specified below is invalid, please generate one entry in keystore only readable by yourself using the encrypt.bat utility -->
                 <entry key="sfdc.password" value="1111111111111111"/>
```
cbhacking: Passwords "encrypted" without a key file will no longer be possible; might be good to make that clear. Alternatively, don't use the per-entry key file path at all; either always use the default secure path, or only set the path in one top-level place.

src/main/java/com/salesforce/dataloader/security/EncryptionAesUtil.java
```diff
@@ -0,0 +1,198 @@
...
+    public byte[] generateEncryptionKey() {
+        // The secure random uses threadId and tick to make sure it is not repeatable
+        SecureRandom random = new SecureRandom();
+        byte bytes[] = new byte[ENCRYPTION_KEY_LENGTH_IN_BYTES];
+        random.nextBytes(bytes);
+        return bytes;
+    }
+
+    public String createUserProfileKeyName() {
+        String path = System.getProperty("user.home") + File.separator + "dataloader";
```
jjangsfdc: `Paths.get(System.getProperty("user.home"), "dataloader").toString()` is what you want. 
Also, the path returned is not consistent with other config files path but I can fix this in my world-accessible file bug(W-4648801)

src/main/java/com/salesforce/dataloader/security/EncryptionAesUtil.java
```diff
@@ -0,0 +1,198 @@
...
+        File customDir = new File(path);
+
+        if (customDir.exists()) {
+            LOGGER.info(customDir + " exists");
+        } else if (customDir.mkdirs()) {
+            LOGGER.info(customDir + " was created");
+        } else {
+            LOGGER.info(customDir + " was not created");
+            throw new RuntimeException("Cannot create keyFile in directory:" + path);
```
jjangsfdc: You mean "cannot create directory"

src/main/java/com/salesforce/dataloader/security/EncryptionAesUtil.java
```diff
@@ -0,0 +1,198 @@
...
+        }
+        return path + File.separator + DEFAULT_KEYFILE_NAME;
```
jjangsfdc: `Paths.get(path, DEFAULT_KEYFILE_NAME).toString()`

src/main/java/com/salesforce/dataloader/security/EncryptionAesUtil.java
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    public String createKeyFile(String filePath) throws IOException {
+        if (filePath == null || filePath.isEmpty()) {
+            filePath = createKeyFile(createUserProfileKeyName());
+        }
+        byte[] key = generateEncryptionKey();
+        try (FileOutputStream fos = new FileOutputStream(filePath)) {
+            fos.write(key);
+        }
```
jjangsfdc: Why not `catch` for IOException within this method instead of bubbling up?

src/main/java/com/salesforce/dataloader/security/EncryptionAesUtil.java
```diff
@@ -0,0 +1,198 @@
...
+            }
+            return new String(cipherKey, "UTF-8");
+        }
+    }
+
+    public String encryptMsg(String msg) throws GeneralSecurityException {
+        try {
+            return EncryptionUtil.bytesToText(encryptMsg(msg, cipherKey));
+        } catch (Exception e) {
+            LOGGER.error("Error to encrypt message: " + e.getMessage());
```
jjangsfdc: Better to say "Failed to encrypt message: "

src/main/java/com/salesforce/dataloader/security/EncryptionAesUtil.java
```diff
@@ -0,0 +1,198 @@
...
+            throw new  GeneralSecurityException("Error to encrypt message: ", e);
+        }
+    }
+
+    public String decryptMsg(String cipherMsgString) throws GeneralSecurityException {
+        try {
+            byte[] cipherMsg = EncryptionUtil.textToBytes(cipherMsgString);
+            return decryptMsg(cipherMsg, cipherKey);
+        } catch (Exception e) {
+            LOGGER.error("Error to decryptMsg: " + e.getMessage());
```
jjangsfdc: "Failed to.."

src/main/java/com/salesforce/dataloader/security/EncryptionUtil.java
```diff
@@ -98,28 +115,27 @@ private static String bytesToText(byte[] bytes) {
...
     }
 
+
     /**
      * Create a key for encryption and decryption
      *
      * @return Key
-     * @throws NoSuchAlgorithmException
-     * @throws InvalidKeySpecException
-     * @throws InvalidKeyException
```
jjangsfdc: You still have these exceptions as throws, but why removed them from comments?

I see that all @throws and @return in comments are removed in this CL. Any particular reason?

src/main/java/com/salesforce/dataloader/security/EncryptionUtil.java
jjangsfdc: Agreed, let's remove this. Leaving in dead code will only make future maintenance harder.

src/main/java/com/salesforce/dataloader/security/EncryptionUtil.java
jjangsfdc: Agreed.

src/main/nsis/config.properties
```diff
@@ -1,5 +0,0 @@
-#Loader Config
```
jjangsfdc: We need this file. Did you delete this intentionally?

src/main/java/com/salesforce/dataloader/config/Config.java
```diff
@@ -791,8 +806,8 @@ public void save() throws IOException, GeneralSecurityException {
 
         // Secure password code prevents the saving of passwords
         // no great way to do this,
-        String oldPassword = encryptProperty(PASSWORD);
-        String oldProxyPassword = encryptProperty(PROXY_PASSWORD);
+        String oldPassword = "";
+        String oldProxyPassword = "";
 
         String oauthAccessToken = getString(OAUTH_ACCESSTOKEN);
         String oauthRefreshToken = getString(OAUTH_REFRESHTOKEN);
```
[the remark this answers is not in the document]
xbiansf: Will update.

pom.xml
```diff
@@ -1,834 +0,0 @@
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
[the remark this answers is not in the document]
xbiansf: Interesting, for some reason, it is deleted.  Then I did not add the pom.xml back to the commit as there is some change I  should not check in.   I will squash it.

src/main/java/com/salesforce/dataloader/config/Config.java
```diff
@@ -195,6 +206,7 @@
     public static final String LOAD_ROW_TO_START_AT = "process.loadRowToStartAt"; //$NON-NLS-1$
     public static final String INITIAL_LAST_RUN_DATE = "process.initialLastRunDate";
     public static final String ENCRYPTION_KEY_FILE = "process.encryptionKeyFile"; //$NON-NLS-1$
+    public static final String ENCRYPTION_KEY_FILE_PASSWORD = "process.encryptionKeyFilePassword"; //$NON-NLS-1$
```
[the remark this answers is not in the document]
xbiansf: Keystore is removed.

src/main/nsis/samples/conf/process-conf.xml
xbiansf: Since we are giving each entry the option of username/password, we can let customer use differnt keyfile as well. Does not hurt us.

src/main/java/com/salesforce/dataloader/security/EncryptionUtil.java
xbiansf: Using throws and returns are not recommended any more.

src/main/java/com/salesforce/dataloader/security/EncryptionAesUtil.java
xbiansf: Catch means you need to handle this failure by using return false to bubble up the error, which is more handling in this case.

src/main/java/com/salesforce/dataloader/security/EncryptionAesUtil.java
```diff
@@ -0,0 +1,195 @@
...
+    public String encryptMsg(String msg) throws GeneralSecurityException {
+        try {
+            return EncryptionUtil.bytesToText(encryptMsg(msg, cipherKey));
+        } catch (Exception e) {
+            LOGGER.error("Fail to encrypt message: " + e.getMessage());
+            throw new  GeneralSecurityException("Error to encrypt message: ", e);
+        }
+    }
+
+    public String decryptMsg(String cipherMsgString) throws GeneralSecurityException {
```
cbhacking: Same as for `encryptMsg(String)` above.

src/main/nsis/samples/conf/process-conf.xml
```diff
@@ -14,7 +14,7 @@
                 <!-- password below has been encrypted using key file, therefore it will not work without the key setting: process.encryptionKeyFile
                 the password is not a valid encrypted value, please generate the real value using encrypt.bat utility -->
                 <entry key="sfdc.password" value="5555555555555555"/>
-                <entry key="process.encryptionKeyFile" value="c:\dataloader\samples\conf\sample.key"/>
+                <entry key="process.encryptionKeyFile" value="c:\users\username\sample.key"/>
```
cbhacking: I recommend suggesting a subdirectory for the keyfile, such as `AppData\Local\SalesforceDataLoader\sample.key`.

src/main/nsis/samples/conf/process-conf.xml
```diff
@@ -34,12 +34,12 @@
         <property name="name" value="opportunityUpsertProcess"/>
         <property name="configOverrideMap">
             <map>
-                <entry key="sfdc.endpoint" value="https://login.salesforce.com"/> 
+                <entry key="sfdc.endpoint" value="https://login.salesforce.com"/>
```
cbhacking: Is it actually necessary to let users set this value? It's a risk; I was able to steal credentials (when the real file is world-writable) but pointing it at a server I control.

src/test/java/com/salesforce/dataloader/EncryptionUtil/EncryptionUtilTest.java
```diff
@@ -0,0 +1,91 @@
...
+
+    @Test /* test for single line encryption */
+    public void testEncryption() throws Exception {
+        EncryptionAesUtil encryptionAesUtil = new EncryptionAesUtil();
+        String passwordText = "somePassword6c3708b3";
+        byte[] secureKey = encryptionAesUtil.generateEncryptionKey();
+        byte[] encryptedMsg = encryptionAesUtil.encryptMsg(passwordText, secureKey);
+        String decryptedMsg = encryptionAesUtil.decryptMsg(encryptedMsg, secureKey);
+        logger.info("\nEncrypted text:" + encryptedMsg.toString() +"\nText to be encrypted:" + passwordText +  "\nDecrypted       text:" + decryptedMsg);
+        Assert.assertEquals("Text recovered from encrypted message is as the expected: ", passwordText, decryptedMsg);
```
cbhacking: Include a negative test. At a minimum, the ciphertext must not match the plaintext.

src/main/java/com/salesforce/dataloader/security/EncryptionAesUtil.java
```diff
@@ -0,0 +1,195 @@
...
+    synchronized public byte[] encryptMsg(String msg, byte[] encryptionKey) throws NoSuchPaddingException, NoSuchAlgorithmException, InvalidKeySpecException, InvalidKeyException, InvalidParameterSpecException, UnsupportedEncodingException, BadPaddingException, IllegalBlockSizeException {
+
+        if(encryptionKey == null)
+            return msg.getBytes("UTF-8");
+        SecretKeySpec key = new SecretKeySpec(encryptionKey, "AES");
+        cipher.init(Cipher.ENCRYPT_MODE, key);
+
+        byte[] cipherText = cipher.doFinal(msg.getBytes("UTF-8"));
+        byte[] iv = cipher.getIV();
+        return concatenateByteArray(iv, cipherText);
```
cbhacking: The length of the automatically-generated IV doesn't seem to be well-defined anywhere. This means that different versions of the JRE, or even just different crypto providers (Java's crypto API is pluggable), may return different generated IV lengths. Currently, this class assumes that the IV length is the same as the key length, but this is by no means guaranteed. A custom IV (use SecureRandom to generate it, same as the key, but *unique for every message*) of a known length would avoid this risk.

src/main/java/com/salesforce/dataloader/security/EncryptionAesUtil.java
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    synchronized public String decryptMsg(byte[] cipherMsg, byte[] encryptionKey) throws InvalidKeySpecException, NoSuchAlgorithmException, BadPaddingException, IllegalBlockSizeException, InvalidAlgorithmParameterException, InvalidKeyException, UnsupportedEncodingException {
```
cbhacking: Same top-level comment as for `encryptMsg(String,byte[])`. This function modifies the global `cipher`, so either it needs to be global or `cipher` needs to be an instance variable.

src/main/java/com/salesforce/dataloader/security/EncryptionAesUtil.java
```diff
@@ -0,0 +1,195 @@
...
+        if (encryptionKey == null)
+            return new String(cipherMsg,"UTF-8");
```
cbhacking: Returning the ciphertext as though it was plaintext is less catastrophic than the reverse, but is still incorrect behavior. It should throw an IllegalArgumentException. Also, trying to decode a byte array that was not explicitly constructed from UTF-8 code points as UTF-8 is going to frequently fail, resulting in a completely bizarre exception being thrown.

src/main/java/com/salesforce/dataloader/security/EncryptionAesUtil.java
```diff
@@ -0,0 +1,195 @@
...
+        } catch (Exception e) {
+            LOGGER.error("Fail to initialize encryption: " + e.getMessage());
+            throw new  RuntimeException("Fail to initialize encryption:  ", e);
+        }
+    }
+
+
+    private byte[] extractIvBytes(byte[] cipheredText) {
+        byte[] iv = new byte[ENCRYPTION_KEY_LENGTH_IN_BYTES];
+        System.arraycopy(cipheredText, 0, iv, 0, ENCRYPTION_KEY_LENGTH_IN_BYTES);
```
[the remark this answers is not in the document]
xbiansf: I am aware of it.  This is only used to defined a constant of 16 byte for AES 128 bit encryption.

src/main/java/com/salesforce/dataloader/security/EncryptionAesUtil.java
```diff
@@ -0,0 +1,195 @@
...
+        try (FileInputStream fis = new FileInputStream(filePath)) {
+            int size = fis.read(data);
+            if (size < ENCRYPTION_KEY_LENGTH_IN_BYTES)
+                throw new GeneralSecurityException("Keyfile content is too short:" + filePath);
+            else
+            {
+                cipherKey = new byte[ENCRYPTION_KEY_LENGTH_IN_BYTES];
+                System.arraycopy(data, 0, cipherKey, 0, ENCRYPTION_KEY_LENGTH_IN_BYTES);
+            }
+            return new String(cipherKey, "UTF-8");
```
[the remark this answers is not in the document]
xbiansf: Was encoding the key to debug and test.   I will remove code to encode the key.

src/main/nsis/samples/conf/process-conf.xml
xbiansf: Yes. We need that.  That is also for custom domain.

src/main/java/com/salesforce/dataloader/security/EncryptionAesUtil.java
xbiansf: Will remove the plaintext option.  Regarding to encoding, as long as they are the same, it will be fine.

src/main/java/com/salesforce/dataloader/security/EncryptionAesUtil.java
```diff
@@ -0,0 +1,255 @@
...
+
+    private byte[] concatenateByteArray(byte[] a, byte[] b) {
+
+        byte[] c = new byte[a.length + b.length];
+        System.arraycopy(a, 0, c, 0, a.length);
+        System.arraycopy(b, 0, c, a.length, b.length);
+        return c;
+    }
+
+    public byte[] generateEncryptionKey() {
```
cbhacking: I would expect this function to always set `cipherKey`, rather than returning an array that should (but might not always) be assigned to `cipherKey`. There's not any reason I can see to ever have key data in any array other than `cipherKey` and it runs the risk of data being encrypted/decrypted with the wrong key, or the wrong key bytes being written to a keyfile.

A separate function that generates an IV could be used for IV; using this function for an IV is arguably incorrect from a code maintainability perspective anyhow (keys are not IVs, even though they are often generated the same way).

src/main/java/com/salesforce/dataloader/security/EncryptionAesUtil.java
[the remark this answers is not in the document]
cbhacking: Ah, you changed `encryptMsg(String, byte[])` to generate an IV that is explicitly this length, that makes sense.

src/main/java/com/salesforce/dataloader/security/EncryptionAesUtil.java
```diff
@@ -0,0 +1,255 @@
...
+        // the first call to this method forces this SecureRandom object to seed itself.
+        // This self-seeding will not occur if setSeed was previously called.
+        SecureRandom random = new SecureRandom();
+        byte bytes[] = new byte[ENCRYPTION_KEY_LENGTH_IN_BYTES];
+        random.nextBytes(bytes);
+        return bytes;
+    }
+
+    public String createUserProfileKeyName() {
+        String path = Paths.get(System.getProperty("user.home"), ".dataloader").toString();
```
cbhacking: Convention on Windows would put this under AppData (probably under the Local subdirectory, though as of writing this the config dirs are stored in Roaming) rather than directly under the profile, but this works.

Just bear in mind that if the config file roams but the keyfile doesn't (and IMO it shouldn't) then the config files will be useless on other machines until they are updated to contain passwords encrypted using keyfiles local to those machines.

src/main/java/com/salesforce/dataloader/security/EncryptionAesUtil.java
```diff
@@ -0,0 +1,255 @@
...
+        File customDir = new File(path);
+
+        if (customDir.exists()) {
+            LOGGER.debug(customDir + " exists");
+        } else if (customDir.mkdirs()) {
+            if (detectedOS == OSType.MacOS || detectedOS == OSType.Linux) {
+                // set all reading to false
+                customDir.setReadable(false, false);
+                // only owner can read
+                customDir.setReadable(true, true);
```
cbhacking: You should remove execute permissions (from everybody other than the owner) from the containing directory as well, as an extra layer of protection for files inside (dirs can't be listed without Read permissions but can be traversed to access their contents - assuming the names are known - unless Execute permissions are also removed).

src/main/java/com/salesforce/dataloader/security/EncryptionAesUtil.java
```diff
@@ -0,0 +1,255 @@
...
+            }
+            LOGGER.info(customDir + " was created");
+
+        } else {
+            LOGGER.info(customDir + " was not created");
```
cbhacking: Shouldn't this be an error (or something else other than `info`) in the log?

src/main/java/com/salesforce/dataloader/security/EncryptionAesUtil.java
```diff
@@ -0,0 +1,255 @@
...
+                cipherKey = new byte[ENCRYPTION_KEY_LENGTH_IN_BYTES];
+                System.arraycopy(data, 0, cipherKey, 0, ENCRYPTION_KEY_LENGTH_IN_BYTES);
+            }
+        } catch (IOException io) {
+            throw new GeneralSecurityException("Failed to open file: " + filePath, io);
+        }
+    }
+
+    private void ensureKeyIsSet() throws IOException, GeneralSecurityException {
+        if (cipherKey != null) {
```
cbhacking: This should check the key for validity more comprehensively (for example, ensuring it's the right length).

src/main/java/com/salesforce/dataloader/security/EncryptionAesUtil.java
```diff
@@ -0,0 +1,255 @@
...
+            return;
+        }
+        // if no key was set up already, create one
+        createKeyFileIfNotExisting(null);
```
cbhacking: This function will set the `cipherKey` to the contents of the keyfile whether or not the keyfile was already existing, unless the keyfile is too short (in which case it won't replace it with a valid one, but will at least throw an exception). However, that series of checks and outcomes is not obvious from the function names (or comments, which are mostly missing); it might be better to make them more explicit

src/main/java/com/salesforce/dataloader/security/EncryptionAesUtil.java
```diff
@@ -0,0 +1,195 @@
...
+ * @author xbian
+ */
+public class EncryptionAesUtil {
+
+    private static final Logger LOGGER = Logger.getLogger(EncryptionAesUtil.class);
+
+    // Support single text encryption and decryption
+
+    static public Cipher cipher;
+    static public byte[] cipherKey;
```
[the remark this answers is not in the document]
cbhacking: Also, why are they public? Other classes should never need to directly touch these variables; if they attempt to do so, it's probably a bug.

src/main/java/com/salesforce/dataloader/security/EncryptionAesUtil.java
```diff
@@ -0,0 +1,255 @@
...
+        byte[] ivBytes = generateEncryptionKey();
+        IvParameterSpec iv = new IvParameterSpec(ivBytes);
+        cipher.init(Cipher.ENCRYPT_MODE, key, iv);
+        byte[] cipherText = cipher.doFinal(msg.getBytes());
+        return concatenateByteArray(ivBytes, cipherText);
+    }
+
+    synchronized public String decryptMsg(byte[] cipherMsg, byte[] encryptionKey) throws GeneralSecurityException, UnsupportedEncodingException {
+        if (encryptionKey == null)
+            throw new GeneralSecurityException("Encryption key is null or has invalid length");
```
cbhacking: This doesn't actually check the length at present.

src/main/java/com/salesforce/dataloader/security/EncryptionAesUtil.java
```diff
@@ -0,0 +1,255 @@
...
+            ensureKeyIsSet();
+            byte[] cipherMsg = EncryptionUtil.textToBytes(cipherMsgString);
+            return decryptMsg(cipherMsg, cipherKey);
+        } catch (Exception e) {
+            LOGGER.error("Fail to decrypt message: " + e.getMessage());
+            throw new GeneralSecurityException("Error to encrypt message: ", e);
+        }
+    }
+
+    synchronized public byte[] encryptMsg(String msg, byte[] encryptionKey) throws GeneralSecurityException, UnsupportedEncodingException {
```
cbhacking: Why is this function `synchronized` and `public`? It's not `static`, so it synchronizes against the instance, but you also said thread safety isn't a goal here in which case synchronizing at all is just misleading (and also wasteful of performance). It's also not clear why both these encrypt/decrypt functions and the versions that don't take an explicit key parameter are `public`.

src/main/nsis/samples/conf/process-conf.xml
cbhacking: This should presumably, at least, be changed to the new default keyfile path...

src/main/java/com/salesforce/dataloader/security/EncryptionAesUtil.java
```diff
@@ -0,0 +1,265 @@
...
+/**
+ * com.salesforce.dataloader.security
+ *
+ * @author xbian
+ */
+public class EncryptionAesUtil {
+
+    private static final Logger LOGGER = Logger.getLogger(EncryptionAesUtil.class);
+
+    public enum OSType {
```
jjangsfdc: This OS related logic can be its own util so that other code can use it too.

PR #152, opened 2018-02-16
Fix sensitive data in world-accessible folders

jjangsfdc:
-Change Windows NSI script to install user specific data in non-world-accessible place
-Change where user's config.properties files are saved
-Add more logging in config bootstrap


pom.xml
```diff
@@ -740,7 +740,6 @@
                         <jvmVersion>1.8</jvmVersion>
                         <jvmOptions>
                             <string>-XstartOnFirstThread</string>
-                            <string>-Dsalesforce.config.dir=Contents/Resources/conf</string>
```
jjangsfdc: Previously, in Mac profile we were ALWAYS passing in "Contents/Resources/conf" as "salesforce.config.dir" and this path is world accessible. We need to dynamically set this directory during the runtime, not during packaging.

pom.xml
```diff
@@ -767,11 +766,11 @@
...
-                            <replace file="${basedir}/target/${project.build.finalName}/${project.name}.app/Contents/Resources/conf/config.properties" token="\n" value="${line.separator}"/>
+                          <echo>Copying additional assets for DMG</echo>
+                          <copy file="${basedir}/src/main/resources/Localizable.strings" tofile="${basedir}/target/${project.build.finalName}/${project.name}.app/Contents/Resources/en.lproj/Localizable.strings"/>
+                          <copy file="${basedir}/src/main/nsis/config.properties" tofile="${basedir}/target/${project.build.finalName}/${project.name}.app/Contents/Resources/conf/defaultConfig.properties"/>
+                          <replace file="${basedir}/target/${project.build.finalName}/${project.name}.app/Contents/Resources/conf/defaultConfig.properties" token="#placeholder" value="${config.additional.properties}"/>
+                          <replace file="${basedir}/target/${project.build.finalName}/${project.name}.app/Contents/Resources/conf/defaultConfig.properties" token="\n" value="${line.separator}"/>
```
jjangsfdc: This `defaultConfig.properties` is used as a default and will be copied to user specific config dir during runtime if they don't have `config.properties` yet.

setup.nsi
```diff
@@ -183,9 +183,9 @@ Section "Required Files"
     File "target\${PROJECT_FINAL_NAME}.exe"
     File "target\${PROJECT_FINAL_NAME}-uber.jar"
     File "src\main\resources\img\icons\dataloader.ico"
+    File "/oname=defaultConfig.properties" "target\config.properties"
```
jjangsfdc: Same as Mac profile above, for Windows we are making a default config to be used when user doesn't have a config yet.

setup.nsi
```diff
@@ -61,7 +61,7 @@ Function InstModeChanged
 SetShellVarContext CURRENT
 ${IfNotThen} ${Silent} ${|} StrCpy $InstDir "${S_DEFINSTDIR_USER}" ${|}
 ${If} $InstMode > 0
-	SetShellVarContext ALL
+	SetShellVarContext CURRENT
```
jjangsfdc: Somehow this `SetShellVarContext ALL` was making %APPDATA% to be in `C:\ProgramData` from `C:\Users\{user}\AppData\Roaming` which we do not want.

setup.nsi
```diff
@@ -183,9 +183,9 @@ Section "Required Files"
     File "target\${PROJECT_FINAL_NAME}.exe"
     File "target\${PROJECT_FINAL_NAME}-uber.jar"
     File "src\main\resources\img\icons\dataloader.ico"
+    File "/oname=defaultConfig.properties" "target\config.properties"
     FileOpen $9 "${PROJECT_FINAL_NAME}.l4j.ini" w
     ;Java Args here
-    FileWrite $9 "-Dappdata.dir=$\"$APPDATA$\"$\r$\n"
```
jjangsfdc: Same as Mac profile above, do not set where `config.properties` will stored during packaging time.

setup.nsi
```diff
@@ -313,6 +313,7 @@ File "/oname=${extractTo}" "${UNINSTEXE}.exe.un"
       Delete "$INSTDIR\${PROJECT_FINAL_NAME}.exe"
       Delete "$INSTDIR\dataloader.ico"
       Delete "$INSTDIR\dataloader_uninstall.exe"
+      Delete "$INSTDIR\defaultConfig.properties"
```
jjangsfdc: Clean up

src/main/java/com/salesforce/dataloader/controller/Controller.java
```diff
@@ -273,6 +284,66 @@ private void copyFile(File file, File destFile) {
...
+                ? Paths.get(ProgramDirectoryUtil.getProgramDirectory())
+                : Paths.get(System.getProperty("user.dir"), "Contents/Resources/conf");
+
+        File file = new File(path.toString());
+        return file;
+    }
+
     /**
      * Get the current config.properties and load it into the config bean.
      *
```
jjangsfdc: Most of the code in this method was a mess. I had to rewrite most of the logic.

src/main/java/com/salesforce/dataloader/controller/Controller.java
```diff
@@ -273,6 +284,66 @@ private void copyFile(File file, File destFile) {
...
+            logger.info("Config directory already exists: " + dirPath);
+        }
+        return isSuccessful;
+    }
+
+    /**
+     * Returns current user's Dataloader configuration directory
+     *  ie) For Windows - C:\Users\{user}\AppData\Roaming\salesforce.com\Data Loader {version}\conf
+     *      For Mac - /Users/{user}/Library/Preferences/salesforce.com/Data Loader {version}/conf
+     *      For Unix* - {user.home}/Library/Preferences/salesforce.com/Data Loader {version}/conf
```
jjangsfdc: These are where sensitive files will be stored

src/main/java/com/salesforce/dataloader/util/ProgramDirectoryUtil.java
```diff
@@ -0,0 +1,74 @@
...
+import java.io.File;
+import java.net.URISyntaxException;
+
+/**
+ * This is an utility class to find out the path of Dataloader
+ *
+ * @author Jae Jang
+ *
+ */
+public class ProgramDirectoryUtil {
```
jjangsfdc: System.getProperty("user.dir") did not always return consistent or correct dir in Windows so this class is used to find the hosted directory of Dataloader

setup.nsi
```diff
@@ -313,6 +313,7 @@ File "/oname=${extractTo}" "${UNINSTEXE}.exe.un"
...
       RMDir /r "$INSTDIR\licenses"
       RMDir /r "$INSTDIR\samples"
       RMDir /r "$INSTDIR\bin"
```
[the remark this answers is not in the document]
jjangsfdc: Yes, there could be another version of DL installed

src/main/java/com/salesforce/dataloader/controller/Controller.java
jjangsfdc: Yes I agree it's not how Unix/Linux would store user specific files.  Knowing that Unix/linux is not officially supported for DL, I didn't cover the case in detail. I will take care of it.

src/main/java/com/salesforce/dataloader/controller/Controller.java
```diff
@@ -273,6 +284,66 @@ private void copyFile(File file, File destFile) {
...
+     *     For Mac - /Applications/Data Loader.app/Contents/Resources/conf
+     *     For Unix* - {user.dir}/Contents/Resources/conf
+     *
+     * @return
+     *  Default Dataloader configuration directory
+     */
+    private static File getDefaultConfigDir() {
+        Path path = IS_WINDOWS
+                ? Paths.get(ProgramDirectoryUtil.getProgramDirectory())
+                : Paths.get(System.getProperty("user.dir"), "Contents/Resources/conf");
```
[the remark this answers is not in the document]
jjangsfdc: > but this path says it'll just find the Contents/... subdir of the current working directory.

Yes you are right. I will revisit this.

> Also, for Linux (or other Unix-likes) this seems to assume that the working directory is wherever the binary was extracted, which is not guaranteed.

Will revisit this too

src/main/java/com/salesforce/dataloader/controller/Controller.java
```diff
@@ -281,91 +352,83 @@ private void copyFile(File file, File destFile) {
...
-                } else {
-                    copyFile(oldConfig, configFile);
-                }
+        String configDirPath = getConfigDir();
+        File configDir;
+        if (IS_WINDOWS) {
+            if ((name == null || name.equals(DataLoaderRunner.UI) == false) && configDirPath == null ) {
+                // Windows commandline and config dir is not provided - use '{user.dir}\target' as config dir
+                logger.debug("Windows commandline and config dir is not provided");
+                configDir = new File(System.getProperty("user.dir"), "target");
```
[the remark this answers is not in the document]
jjangsfdc: AFAIK, DL in command line mode is not officially supported for Unix* or Mac. It is only supported in Windows, but I know some customers were able to tweak things and made CLI work in Linux too. I wanted to merely change the config dir location logic but still keep the original logic.

I do see your point of having consistent behavior across all platforms. I will see what I can do.

src/main/java/com/salesforce/dataloader/controller/Controller.java
```diff
@@ -281,91 +352,83 @@ private void copyFile(File file, File destFile) {
...
+        appPath = configDir.getAbsolutePath();
+
+        // check if the config file exists
+        File configFile = new File(appPath, CONFIG_FILE);
+        String configPath = configFile.getAbsolutePath();
+        if (!configFile.exists()) {
+
+            // If config file doesn't exist, copy default config
+            File defaultConfigFile = new File(getDefaultConfigDir(), DEFAULT_CONFIG_FILE);
+            copyFile(defaultConfigFile, configFile);
```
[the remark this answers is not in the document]
jjangsfdc: Good question. It was there from the beginning and not sure what's special about it. I will see.

src/main/java/com/salesforce/dataloader/controller/Controller.java
```diff
@@ -376,15 +439,15 @@ protected void initConfig(String name, boolean isBatchMode) throws ControllerIni
...
+    private static String getProductName() {
         return APP_NAME + " " + APP_VERSION;
     }
 
     public static synchronized void initLog() throws FactoryConfigurationError {
         // init the log if not initialized already
         if (Controller.isLogInitialized) { return; }
         try {
-            File logConfXml = new File(System.getProperty("user.dir"), LOG_CONF_OVERRIDE);
+            File logConfXml = new File(getDefaultConfigDir(), LOG_CONF_OVERRIDE);
```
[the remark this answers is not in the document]
jjangsfdc: I thought about this and concluded that log setting might be something it should be set across for all users instead of each user controlling it. The log file is persisted in system-wide location.

I just realized that the log file is persisted in system-wide location in Mac but not in Windows.
Windows - `C:\Users\{user}\AppData\Local\Temp\`
Mac -  `/var/folders/…`

Do you think logs should be stored in user-specific or system-wide location?

src/main/java/com/salesforce/dataloader/util/ProgramDirectoryUtil.java
jjangsfdc: Right now I am only calling getProgramDirectory() for Windows. Would it be safer to call getProgramDirectory() for any OS then?

src/main/java/com/salesforce/dataloader/util/ProgramDirectoryUtil.java
```diff
@@ -0,0 +1,74 @@
...
+                .getLocation()
+                .getPath())
+                .getName();
+
+        return jarName.contains(".jar");
+    }
+
+    private static String getCurrentProjectDirectory()
+    {
+        return new File("").getAbsolutePath();
```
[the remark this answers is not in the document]
jjangsfdc: Yes that is it.

src/main/java/com/salesforce/dataloader/controller/Controller.java
```diff
@@ -243,47 +242,6 @@ public synchronized boolean saveConfig() {
 
     }
 
-    /**
```
jjangsfdc: Removed unnecessary custom method.

src/main/java/com/salesforce/dataloader/controller/Controller.java
```diff
@@ -357,40 +340,16 @@ protected void initConfig(String name, boolean isBatchMode) throws ControllerIni
...
-            else if (name != null && name.equals("ui")) {
-                // Windows UI - use user's config dir
-                logger.debug("Windows UI");
-                configDir= getUserConfigDir();
-            }
-            else {
-                logger.error("Unknown application mode!");
-                throw new ControllerInitializationException("Unknown application mode!");
-            }
+
```
jjangsfdc: The logic is now simple. Whether you have config dir provided or not.

src/main/java/com/salesforce/dataloader/controller/Controller.java
```diff
@@ -405,13 +364,38 @@ else if (name != null && name.equals("ui")) {
...
+                    logger.info(String.format("User config file does not exist in '%s' Default config file is copied from '%s'",
+                            configPath, defaultConfigFile.getAbsolutePath()));
+                    Files.copy(defaultConfigFile.toPath(), configFile.toPath(), StandardCopyOption.REPLACE_EXISTING);
+                } catch (IOException e) {
+                    String errorMsg = String.format("Failed to copy '%s' to '%s'", defaultConfigFile.getAbsolutePath(), configFile);
+                    logger.warn(errorMsg, e);
+                    throw new ControllerInitializationException(errorMsg, e);
+                }
+            }
+            else {
```
jjangsfdc: This is the case when Dataloader is launched in dev mode

setup.nsi
cbhacking: That's not the typical meaning of that option. Generally, the only reason you'd install software for the "current user only" is if you didn't have admin privileges. "Install for all users" usually just means the following:

* Install in a world-readable-but-only-admin-writable location (usually `Program Files`), rather than in a single-user-R/W-access location (usually the user's profile directory).
* Put the uninstall info and any machine-wide configuration data in the `HKEY_LOCAL_MACHINE` root registry key, rather than in the `HKEY_CURRENT_USER` key.
* Put Start Menu shortcuts under the All Users Start Menu, rather than the Current User Start menu.

Actual application data, generated by or for individual users, should always be stored in the user's profile (as is now being done). `All Users`/`ProgramData`, like `HKEY_LOCAL_MACHINE` is a suitable location for things like license information (if the app is licensed on a per-install basis), configuration options an administrator wants to enforce for all users, and so on.

src/main/java/com/salesforce/dataloader/controller/Controller.java
```diff
@@ -234,43 +243,88 @@ public synchronized boolean saveConfig() {
...
+     *     For *nix - {where DL is installed}/conf
+     *
+     * @return
+     *  Default Dataloader configuration directory
+     */
+    private static File getDefaultConfigDir() {
+        File dir;
+        switch (OS_TYPE) {
+            case WINDOWS: {
+                dir = new File(ProgramDirectoryUtil.getProgramDirectory(), "conf");
```
cbhacking: The result of `getProgramDirectory` is not guaranteed to be an install directory (that has a `conf` subdir). Notably, it could be the working directory, if the app is not being run from a .jar file.

If getting the install location is intended to be the result of the function, it should be changed to ensure that behavior.

src/main/java/com/salesforce/dataloader/controller/Controller.java
```diff
@@ -281,91 +335,84 @@ private void copyFile(File file, File destFile) {
...
-                configFile.setWritable(true);
-                configFile.setReadable(true);
-                logger.info("config file created at " + configFile.getAbsolutePath());
-            } catch (IOException e) {
-                throw new ControllerInitializationException(e);
+            else {
+                try {
+                    // Create a blank user config in "./target"
+                    appPath = new File(ProgramDirectoryUtil.getProgramDirectory(), "target").getAbsolutePath();
+                    configFile = new File(appPath, CONFIG_FILE);
```
cbhacking: If we get here, then a "proper" config dir exists (possibly having just been created, but still) under the user profile. Given that, why not use it? Why create a new "target" subdirectory of either the install location (which may just fail with access denied) or the working directory (which seems to be what you intend to do, but isn't always what `getProgramDirectory` returns), when you've got a config directory ready to go?

src/main/java/com/salesforce/dataloader/controller/Controller.java
```diff
@@ -234,43 +243,88 @@ public synchronized boolean saveConfig() {
...
+     */
+    private static File getDefaultConfigDir() {
+        File dir;
+        switch (OS_TYPE) {
+            case WINDOWS: {
+                dir = new File(ProgramDirectoryUtil.getProgramDirectory(), "conf");
+                break;
+            }
+            case MACOSX: {
+                dir = new File(System.getProperty("user.dir"), "Contents/Resources/conf");
```
cbhacking: The working directory is not guaranteed to be the install directory. This applies to all platforms.

src/main/java/com/salesforce/dataloader/util/ProgramDirectoryUtil.java
```diff
@@ -0,0 +1,73 @@
...
+ * This is an utility class to find out the path of Dataloader
+ *
+ * @author Jae Jang
+ *
+ */
+public class ProgramDirectoryUtil {
+
+    public static String getProgramDirectory()
+    {
+        return runningFromJAR() ? getCurrentJARDirectory() : System.getProperty("user.dir");
```
cbhacking: What is this function actually supposed to do? It seems like it's sometimes supposed to return the install location, and sometimes the working directory, which is probably not actually the desired behavior. At least, some of the calls to it are likely to fail or behave in a probably-incorrect manner under one or the other of those results. As the function is undocumented, it's not clear what its intent is, either.

src/main/java/com/salesforce/dataloader/controller/Controller.java
```diff
@@ -281,91 +335,84 @@ private void copyFile(File file, File destFile) {
...
+
+        // check if the config file exists
+        File configFile = new File(appPath, CONFIG_FILE);
+        String configPath = configFile.getAbsolutePath();
+        if (!configFile.exists()) {
+
+            File defaultConfigFile = new File(getDefaultConfigDir(), DEFAULT_CONFIG_FILE);
+
+            // If default config exists, copy the default to user config
+            // If doesn't exist, create a blank user config in "./target"
```
cbhacking: Why not use the user config directory?

src/main/java/com/salesforce/dataloader/controller/Controller.java
```diff
@@ -234,43 +243,88 @@ public synchronized boolean saveConfig() {
...
+     *  Current user's Dataloader configuration directory
+     */
+    private static File getUserConfigDir() {
+        File dir;
 
-            in = new BufferedInputStream(new FileInputStream(file), bufferSize);
-            int r = 0;
+        switch (OS_TYPE) {
+            case WINDOWS: {
+                dir = Paths.get(System.getProperty("user.home"), "AppData/Local", APP_VENDOR, getProductName(), CONFIG_DIR).toFile();
```
cbhacking: While this will most likely work (Win32 tries to be compatible when it can, and Java on Windows benefits from that), you're using the wrong character for a Windows path separator here.

setup.nsi
jjangsfdc: Changed it back

src/main/java/com/salesforce/dataloader/controller/Controller.java
jjangsfdc: No longer using `getProgramDirectory`

src/main/java/com/salesforce/dataloader/controller/Controller.java
jjangsfdc: That's okay. If it can't find a default config, it will just create a new config

src/main/java/com/salesforce/dataloader/controller/Controller.java
```diff
@@ -234,43 +241,87 @@ public synchronized boolean saveConfig() {
...
+        return dir;
+    }
 
-            while ((r = in.read(buf, 0, buf.length)) != -1) {
-                out.write(buf, 0, r);
+    /**
+     * Returns default Dataloader configuration directory
+     * ie) For Windows - C:\Program Files (x86)\salesforce.com\Data Loader\conf
+     *     For Mac - /Applications/Data Loader.app/Contents/Resources/conf
+     *     For *nix - {where DL is installed}/conf
```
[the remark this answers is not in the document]
xbiansf: Will remove the comments for Jae.

src/main/java/com/salesforce/dataloader/controller/Controller.java
```diff
@@ -234,43 +241,87 @@ public synchronized boolean saveConfig() {
...
+     *
+     * @return
+     *  Default Dataloader configuration directory
+     */
+    private static File getDefaultConfigDir() {
+        File dir;
+        switch (OS_TYPE) {
+            case WINDOWS: {
+                dir = new File(System.getProperty("user.dir"), "conf");
```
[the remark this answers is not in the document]
xbiansf: The intention is used to get the configuration under user directory.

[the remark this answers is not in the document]
cbhacking: Pretty sure that's incorrect. You're thinking of `getUserConfigDir` (see Controller.java:292); the `getDefaultConfigDir` function is supposed to return the dir containing the default config template, which is under the user directory (see pom.xml:771-773 and associated comment, setup.nsi:195-196,313,317, and Controller.java:88,375-390 for usage).

Again, this isn't a security bug, it just won't (reliably) work the way it's supposed to. Also, the new comment is still misleading: `System.getProperty("user.dir")` returns the working directory, which is (usually) not the user's (home) directory; it might be the install location (which is the only case under which this code will work), the user's DataLoader config directory (as set by the Start menu shortcut on Windows, see setup.nsi:7,201), or wherever the user's shell has as its current working directory (could be anywhere).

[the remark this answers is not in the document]
xbiansf: Agree that was wrong.  Will test: String path = Controller.class.getProtectionDomain()
                .getCodeSource().getLocation().getPath();   and verify if it works.

src/main/java/com/salesforce/dataloader/controller/Controller.java
```diff
@@ -225,7 +234,7 @@ public void createAndShowGUI() throws ControllerInitializationException {
             String errMsg = Messages.getFormattedString("Controller.errorConfigWritable", config.getFilename());
 
             String currentWorkingDir = System.getProperty("user.dir");
-            if (currentWorkingDir.startsWith("/Volumes")){
+            if (currentWorkingDir.startsWith("/Volumes")) {
```
cbhacking: The working dir is not the install dir; this will not reliably catch the user running from a DMG (you'd need to check the install dir for that. See `getInstalledConfigDir` for an example.

src/main/java/com/salesforce/dataloader/controller/Controller.java
```diff
@@ -310,38 +319,59 @@ private static File getUserConfigDir() {
...
+            logger.error("Can find binary's location", e);
+            throw new RuntimeException(e);
+        }
+
         File dir;
         switch (OS_TYPE) {
             case WINDOWS: {
-                dir = new File(System.getProperty("user.dir"), "conf");
+                //For windows,﻿filepath is﻿C:\Program Files\salesforce.com\Data Loader\dataloader-xxx-uber.jar
+                dir = getInstalledConfigDir(path, path.lastIndexOf(File.separator),  "conf");
```
cbhacking: Java provides path traversal and cracking functions. It is generally better to use those than to do it yourself using `lastIndexOf` or `indexOf` and `substring`. However, this does seem to be done correctly here.

src/main/java/com/salesforce/dataloader/controller/Controller.java
```diff
@@ -355,29 +385,33 @@ protected void initConfig(String name, boolean isBatchMode) throws ControllerIni
...
         File configFile = new File(appPath, CONFIG_FILE);
+
         String configPath = configFile.getAbsolutePath();
+        logger.info("Looking for file in config path: " + configPath);
         if (!configFile.exists()) {
 
-            File defaultConfigFile = new File(getDefaultConfigDir(), DEFAULT_CONFIG_FILE);
-
+            File defaultConfigFile = new File(getInstalledConfigDir(), DEFAULT_CONFIG_FILE);
+            logger.debug("Looking for file in config file " + defaultConfigFile.getAbsolutePath());
```
cbhacking: It occurs to me that some amount of naming confusion might be avoidable by calling the "defaultConfig.properties" file the "template" rather than the "default". For example, "templateConfigDir" is a bit less confusing than "defaultConfigDir" was. "installedConfigDir" works, though.

src/main/java/com/salesforce/dataloader/controller/Controller.java
xbiansf: It is a little bit tricky for different OS here.

PR #153, opened 2018-02-22
Clean up OS lookup code and few others

jjangsfdc:
@xbiansf take a look.

### DIFF
--- a/src/main/java/com/salesforce/dataloader/security/EncryptionAesUtil.java
+++ b/src/main/java/com/salesforce/dataloader/security/EncryptionAesUtil.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2015, salesforce.com, inc.
+ * Copyright (c) 2018, salesforce.com, inc.
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without modification, are permitted provided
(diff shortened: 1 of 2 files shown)


src/main/java/com/salesforce/dataloader/security/EncryptionUtil.java
```diff
@@ -75,12 +76,13 @@ public static String bytesToText(byte[] bytes) {
 
 
     private static void printUsage() {
-        LOGGER.info("\nUtility to encrypt a string based on a static or a provided key");
```
jjangsfdc: Why log this? Change it to console out.

src/main/java/com/salesforce/dataloader/security/EncryptionAesUtil.java
```diff
@@ -93,7 +67,7 @@ public static OSType getOperatingSystemType() {
     static {
         try {
             cipher = Cipher.getInstance("AES/CBC/PKCS5Padding");
-            detectedOS = getOperatingSystemType();
+            detectedOS = OSInfo.getOSType();
```
jjangsfdc: We don't have to hand roll a custom method to find OS type.

src/main/java/com/salesforce/dataloader/security/EncryptionUtil.java
```diff
@@ -75,12 +76,13 @@ public static String bytesToText(byte[] bytes) {
...
-                + "\t-d <encryptText> <Path to keyfile>               Decrypt an encrypted text back to plain text value using keyfile\n)"
-                + "\t-k [Path to Keyfile]                             Generate keyfile with optional keyfile path\n"
-        );
+        String usage = "\nUtility to encrypt a string based on a static or a provided key\n"
+                        + "Options (mutually exclusive - use one at a time): \n"
+                        + "\t-e <plain text> <Path to keyfile>                Encrypt a plain text value, keyfile path\n"
+                        + "\t-d <encryptText> <Path to keyfile>               Decrypt an encrypted text back to plain text value using keyfile\n"
+                        + "\t-k [Path to Keyfile]                             Generate keyfile with optional keyfile path\n";
+
+        System.out.println(usage);
```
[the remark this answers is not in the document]
jjangsfdc: DL is printing out the log lines in console because our default log4j appender is set to to stdout but if user changes the log level to something less from INFO like (WARN, ERROR or FATAL), that usage will not show in the console. Whereas, System.out will write out to console regardless of log level

PR #154, opened 2018-02-23
Support only https in dataloader

xbiansf:


src/main/java/com/salesforce/dataloader/ui/UIUtils.java
```diff
@@ -26,27 +26,39 @@
...
-    public static URL newURL(String url_name) {
+    public static boolean isValidHttpsUrl(String url) {
         try {
-            return new URL(url_name);
-        } catch (MalformedURLException e) {
-            throw new RuntimeException("Malformed URL " + url_name, e);
+            // check if it is a valid url
+            new URL(url).toURI();
+            // check if it is https protocol
+            return url.matches("^https://.*$");
```
jjangsfdc: This will not match casing.
It would be better to use this [Uri](https://docs.oracle.com/javase/7/docs/api/java/net/URI.html#getScheme())

src/main/resources/labels.properties
```diff
@@ -71,6 +71,7 @@ AdvancedSettingsDialog.assignmentRule=Assignment rule:
 AdvancedSettingsDialog.externalIdField=External ID field:
 AdvancedSettingsDialog.useEuropeanDateFormat=Use European date format (dd/mm/yyyy)
 AdvancedSettingsDialog.serverURL=Server host (clear for default):
+AdvancedSettingsDialog.serverURLInfo=Server host field only supports HTTPS protocol for security reasons. \"{0}\" is invalid.
```
jjangsfdc: This label is user facing and we should run by doc writer.

src/test/java/com/salesforce/dataloader/ui/UiUtilTest.java
```diff
@@ -0,0 +1,45 @@
+package com.salesforce.dataloader.ui;
+
+/*
+ * Copyright (c) 2015, salesforce.com, inc.
```
jjangsfdc: 2018

src/main/java/com/salesforce/dataloader/ui/UIUtils.java
xbiansf: Sounds better.

PR #156, opened 2018-03-07
Polishing encryption utils prompt

xbiansf:
Fixing some prompt issue.   And change all encryption util to use console out and leave aes util to be using logger because main program will call the AesUtil functions.

### DIFF
--- a/src/main/java/com/salesforce/dataloader/security/EncryptionUtil.java
+++ b/src/main/java/com/salesforce/dataloader/security/EncryptionUtil.java
@@ -25,14 +25,9 @@
  */
 package com.salesforce.dataloader.security;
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/salesforce/dataloader/security/EncryptionUtil.java
```diff
@@ -96,12 +91,12 @@ public static void main(String[] args) {
         int i = 0;
         String option = args[i];
         if (option.length() < 2 || option.charAt(0) != '-') {
-            LOGGER.info("Invalid option format: " + args[i]);
```
neetikumar: So not logging anymore?

PR #159, opened 2018-04-19
Initialize static variables for command line mode

xbiansf:
We added new static variable during security fixes.  But in command line mode, they are initialized correctly.   And also bump the to 42.1.0 for release.

### DIFF
--- a/src/main/java/com/salesforce/dataloader/controller/Controller.java
+++ b/src/main/java/com/salesforce/dataloader/controller/Controller.java
@@ -116,10 +116,21 @@ public class Controller {
     private String appPath;
 
     private Controller(String name, boolean isBatchMode) throws ControllerInitializationException {
-        // load app version properties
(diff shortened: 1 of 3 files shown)


src/main/java/com/salesforce/dataloader/controller/Controller.java
```diff
@@ -299,8 +303,14 @@ private static boolean createDir(File dirPath) {
      * @return Current user's Dataloader configuration directory
      */
     private static File getUserConfigDir() {
+        if (OS_TYPE == null) {
+            logger.error("getUserConfigDir(): Control static values are not initiallized " );
+            throw new RuntimeException("Os type not initialled correctly!");
```
jjangsfdc: Typo

xbiansf: Good catch.  Updated.

PR #166, opened 2018-11-08
Allow to import surrogate pair characters

kenichi-ando:


ashitsalesforce:
Hi Kenichi,

Thanks for the contribution.

PR #171, opened 2018-12-11
Delete swt dll from user space directory

xbiansf:


src/main/java/com/salesforce/dataloader/controller/Controller.java
```diff
@@ -388,6 +406,9 @@ protected void initConfig(String name, boolean isBatchMode) throws ControllerIni
         // Initialize log first to use correct logging level
         initLog();
 
+        // Todo: Remove this after eclipse fixed the SWT issue.
+        deleteSwtDirectory();
```
jjangsfdc: Not checking for OS type?  What if there's actually a file in Mac that user created?

xbiansf: inside the seleteSwtDirectory(), you can see we do not do anything for other OS types. Easier to delete in the future.

PR #172, opened 2019-02-11
Upgrading to V45 force Apis with plugin fixes.

xbiansf:


xbiansf:
Upgrade with only API upgrade and needed pom change to ship with V45 separately.  This is only for Java 8.   Openjdk 11 related changes are not in.

jjangsfdc:
Looks good

PR #176, opened 2019-03-02
Adds duplicate bypass header.

jmdohn:
Adds the ability to bypass error message when adding duplicate to salesforce.  This is useful when fuzzy logic is utilized in matching rules rather than exact matches.


jmdohn:
Closing to merge in with pull rq with pk chunking.

PR #177, opened 2019-03-06
Adds duplicate bypass header, PK Chunking

jmdohn:
Adds ability to bypass duplicate rules (I.E. Useful for fuzzy matching blocks insertion / update).
Adds ability to PK chunk, set start row and set chunk size.


ashitsalesforce:
Hi @jmdohn , I am closing this PR as it has conflicts with the code. Most of the proposed changes are incorporated that should enable a relatively smaller code change to support PK chunking for bulk query when we decide to support this feature enhancement.

PR #199, opened 2020-01-13
Modified dataloader.bat to located vendors JDK

ekgcorp:
Previous versions of the Vendor's JDK install will either default to %programfiles%\JDK1.8xxx or JDK8. However, this is not full proof if the users has modified the InstallDir when it was installed. A better method might be to check the Registry for Computer\HKEY_LOCAL_MACHINE\SOFTWARE\JavaSoft\JDK value for: CurrentVersion then Computer\HKEY_LOCAL_MACHINE\SOFTWARE\JavaSoft\JDK\%CurrentVersion% then get the value for: JavaHome


ashitsalesforce:
PR #247 addresses the issue by using JAVA_HOME env variable to add %JAVA_HOME%\bin to the PATH env variable on Windows.

PR #200, opened 2020-01-22
Suggested update since it's not obvious how to install on windows.

OakNinja:
It's far from obvious that ZULU_JAVA_HOME needs to be set. 

I've added a suggested update to dataloader.bat instructions.


ashitsalesforce:
Thanks for the contribution.

PR #202, opened 2020-01-29
Increase maximum number of proxy port digits in settings dialog

brytonpilling:
Some proxies use port numbers with greater than 4 digits. The settings dialog in Dataloader currently only accepts a maximum of 4 digits for the port number. This change increases the maximum to 5 digits and also increases the length of the text field appropriately.


ashitsalesforce:
Thanks for the contribution.

PR #203, opened 2020-02-10
New Exe, and JDK 11 logic

ekgcorp:
In Previous Pull Request, I added logic to dataloader.bat to check for JDK 11 instead of Zulu. Then I ended up changing it to dataloader.ps1 file to add Popup Message on Error if JDK could not be found. Then I could not get Install.bat to created the shortcut correctly while pointing to Powershell Script. So I compiled it into EXE and changed Reference within install.bat file to use EXE instead.

Finally, I deleted dataloader.bat file because its no longer being used.


ashitsalesforce:
Hi ekgcorp,

We are continuing with the bat file approach for now. Consequently rejecting the PR as we do not want to replace the bat file with a .exe.

PR #212, opened 2020-08-03
Removed spammy 'no value for column' message.

triestelaporte:
I've seen a few complaints about this message showing up during command line bulk operations.  I can't see much value in it.. the message only appears during bulk loads, so it doesn't appear that it was important for non-bulk loads.  It's simple to remove.  I'm happy to add better commentary or anything else needed to adhere to standards.

Thanks!


ashitsalesforce:
How about reducing the log level for the message to debug?

ashitsalesforce:
PR #246 addresses the issue.

PR #220, opened 2020-10-07
pom.xml updates

ashitsalesforce:
- speed up jar building by skipping dmg creation.

- sign jar using mvn target and jarsigner plug-in.

- fix mvn warning about not specifying the version for maven-compiler-plugin.

- include Windows-specific SWT DLLs in the jar if Windows build target is specified.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -20,11 +20,13 @@
     <minJvmVersion>1.11.0</minJvmVersion>
     <codesign>DONTSIGN</codesign>
     <jacob.version>1.14.3</jacob.version>
+    <make.dmg.skip>true</make.dmg.skip>
     <autoit.dll.path>autoitplaceholder</autoit.dll.path>
     <nsis.version>2.46</nsis.version>
     <spring.version>3.1.1.RELEASE</spring.version>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <config.additional.properties></config.additional.properties>
+    <jarsigner.skip>true</jarsigner.skip>
 
     <!-- test properties -->
     <test.endpoint>http://testendpoint</test.endpoint>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -221,6 +223,13 @@
         <directory>src/main/nsis/bin</directory>
         <filtering>true</filtering>
       </resource>
+      <resource>
+        <directory>${DLLFolder}</directory>
+        <filtering>false</filtering>
```
jjangsfdc, line 228: What does this filtering=false do?

pom.xml
```diff
@@ -401,11 +411,29 @@
...
+          <execution>
+            <id>sign</id>
+            <goals>
+              <goal>sign</goal>
+            </goals>
+          </execution>
+        </executions>
+        <configuration>
+          <archive>target/dataloader-${project.version}-uber.jar</archive>
+          <skip>${jarsigner.skip}</skip>
```
jjangsfdc, line 434: Looking at line 29, the variable is set to 'false' then will it not sign the Jar? Also, how will this work for users who would want to build it on their machine without actually signing?

jjangsfdc (commented):
As per yesterday's discussion, did you also incorporated the fix of changing one place to update the whole version?

pom.xml
ashitsalesforce: From https://books.sonatype.com/mvnref-book/reference/resource-filtering-sect-description.html:

Using Maven resource filtering you can reference Maven properties and then use Maven profiles to define different configuration values for different target deployment environments. 

Given that we are not referencing any Maven properties for including DLLs for Windows environment, I set <filtering> to false.

pom.xml
ashitsalesforce: The variable is set to true in pom.xml, thereby skipping jar signing by default. So, anyone who wants to build without signing should run "mvn clean package -DskipTests".

PR #222, opened 2020-10-10
templatize version information and specify the values in pom.xml

ashitsalesforce:
Use existing variable, "version", and introduce a new variable, "short.version", in pom.xml. Use these variables to set values in mac and windows-specific files after copying them to target directory.


jjangsfdc (approved):
LGTM

PR #229, opened 2020-11-20
Update SWT to support MacOS Big Sur

ashitsalesforce:
Update SWT to support MacOS Big Sur


jjangsfdc (approved):
Look good to me. Let's double check to make sure it runs in Windows as well.

ashitsalesforce:
SWT update to support MacOS Big Sur.

PR #233, opened 2020-11-28
Improve usability by throwing error messages

ashitsalesforce:
Stop processing CSV file instead of encountering an NPE if empty header column encountered.

Show error message instead of throwing IndexOutOfBoundsException if there are fewer columns in a data row than header row.


src/main/java/com/salesforce/dataloader/ui/DataSelectionDialog.java
```diff
@@ -147,9 +147,9 @@ public void run() {
                         dataReader.checkConnection();
                         dataReader.open();
 
-                        String warning = DAORowUtil.validateColumns(dataReader);
-                        if(warning != null && warning.length() != 0) {
-                            int response = UIUtils.warningConfMessageBox(shell, warning + "\n" + Labels.getString("DataSelectionDialog.warningConf"));
+                        String error = DAORowUtil.validateColumns(dataReader);
+                        if(error != null && error.length() != 0) {
+                            int response = UIUtils.errorMessageBox(shell, error);
```
jjangsfdc, line 152: This will now show an error modal instead of a warning modal. Does that `errorMessageBox` also have the same prompt buttons for the users? ie. Yes|NO, Okay|Cancel etc...
Wanted to make sure it doesn't break any existing work flow.

ashitsalesforce: That's correct. The error modal will stop the user from going forward because it has only one button "OK" which stops the user from proceeding further (unlike the warning modal). Doing so avoids the NPE that users currently encounter. The NPE occurs because the code tries to insert a cell value in a hashmap with null as the key value (the value of the header).

jjangsfdc (approved):
LGTM. Approved.

PR #235, opened 2020-12-16
Updates for v51

ashitsalesforce:
Updates for v51:
- Bump up the version number in pom.xml.
- dlbuilder.sh a simple shell script that invokes mvn directives to build the dataloader zip files.
- Upgrade Spring from v3.1.1 to v5.2.7.
- Upgrade log4j from 1.2.17 to 2.14.0. Upgrade affected files to use LogManager.getLogger(...) and also update Controller.initLog() to initialize logging.


src/main/resources/templates/process-conf.xml
```diff
@@ -3,7 +3,7 @@
     <!-- For the absolute paths below, replace with a better mechanism, eg. config.properties param for base location -->
     <bean id="maximumBatchRowsDbProcess"
           class="com.salesforce.dataloader.process.ProcessRunner"
-          singleton="false">
+          scope="prototype">
```
jjangsfdc, line 6: Could you explain about the changes?

src/test/resources/testfiles/conf/database-conf.xml
```diff
@@ -2,50 +2,50 @@
 <beans>
 <bean id="testDbDataSource"
       class="org.apache.commons.dbcp.BasicDataSource"
-      destroy-method="close">
+      destroy-method="close" scope="singleton">
```
jjangsfdc, line 5: Could you explain about the changes?

dlbuilder.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh -f
```
jjangsfdc, line 1: Nice 👍

jjangsfdc (commented):
Which IDE are you using? If you were using Intellij, it will help you with 'import' clean up.

src/main/java/com/salesforce/dataloader/action/AbstractAction.java
```diff
@@ -70,7 +70,7 @@
 
     protected AbstractAction(Controller controller, ILoaderProgress monitor)
             throws DataAccessObjectInitializationException {
-        this.logger = Logger.getLogger(getClass());
+        this.logger = org.apache.logging.log4j.LogManager.getLogger(getClass());
```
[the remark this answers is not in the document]
ashitsalesforce: Good call. I will withdraw this PR and resubmit with the change.

src/main/java/com/salesforce/dataloader/controller/Controller.java
```diff
@@ -82,7 +84,8 @@
 public class Controller {
 
     private static final String LOG_CONF_OVERRIDE = "log-conf.xml";
-    private static boolean isLogInitialized = false; // make sure log is initialized only once
+
+	private static boolean isLogInitialized = false; // make sure log is initialized only once
```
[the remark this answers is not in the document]
ashitsalesforce: Good catch. vi editor + use of tab created the issue. I will fix it and resubmit the PR.

src/main/java/com/salesforce/dataloader/controller/Controller.java
```diff
@@ -51,8 +51,10 @@
 import com.sforce.soap.partner.DescribeSObjectResult;
 import com.sforce.ws.ConnectionException;
 
-import org.apache.log4j.Logger;
-import org.apache.log4j.xml.DOMConfigurator;
+import org.apache.logging.log4j.Logger;
```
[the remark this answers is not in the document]
ashitsalesforce: I removed the old ones - org.apache.log4j.Logger and org.apache.log4j.xml.DOMConfigurator as they are not supported.

Rest of them are used, at least per Eclipse IDE.

src/main/java/com/salesforce/dataloader/controller/Controller.java
```diff
@@ -462,24 +465,24 @@ public static synchronized void initLog() throws FactoryConfigurationError {
...
-            logger.info(Messages.getString("Controller.logInit")); //$NON-NLS-1$
-        } catch (Exception ex) {
-            logger.error(Messages.getString("Controller.errorLogInit")); //$NON-NLS-1$
-            logger.error(ex);
-            System.exit(1);
+        File logConfXml = Paths.get(System.getProperty("user.dir"), "configs", LOG_CONF_OVERRIDE).toFile();
+        if (logConfXml.exists()) {
+        	System.setProperty("log4j.configurationFile", logConfXml.getAbsolutePath());
+        } else {
+            System.out.println("Using built-in logging configuration, no log-conf.xml in " + logConfXml.getAbsolutePath());
```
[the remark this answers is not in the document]
ashitsalesforce: I removed the else clause because the logger is not initialize at this point. A check after initialization of the logger makes the following call if logger wasn't initialized from the config file:
if (logConfigLocation == null) {
        	logger.error("Unable to initialize logging using log4j2 config file at "
        			+ logConfXml.getAbsolutePath()
        			+ ". All error messages will be logged on STDOUT.");
        } else {
        	logger.info("Using log4j2 configuration file at location: " + logConfigLocation);
        }

src/main/resources/templates/process-conf.xml
ashitsalesforce: Spring framework upgrade necessitated this change. singleton="false" is not supported in the upgraded version of Spring. Replacement attribute is scope with value "prototype" per https://stackoverflow.com/questions/49879668/singleton-false-is-no-longer-supported-in-spring-4-3-8-need-solution

src/test/resources/testfiles/conf/database-conf.xml
ashitsalesforce: The default Spring bean scope is singleton (i.e. if scope attribute is not explicitly set) per http://blog.marcnuri.com/spring-bean-scopes-guide-to-understand-the-different-bean-scopes/. Given that all other beans in the file had scope set, I added this line during Spring framework upgrade to make the scope explicit and align with its explicit setting for other beans in the same file.

dlbuilder.sh
ashitsalesforce: thanks.

ashitsalesforce:
Withdrawing the PR to make changes based on the feedback.

PR #238, opened 2021-01-06
pom.xml cleanup and related changes

ashitsalesforce:
Cleanup of pom.xml to improve maintainability. Changes include:
- remove unused obsolete profiles, plugin repos, and dependencies
- rearrange dependencies to group the depencies for testing scope together.
- update quartz scheduler and jface to the latest version.
- use local repo for SWT to be able to download the latest jar and use it instead of having to await its availability through a public maven repo, whose version numbers do not match with SWT's version numbers.
- add profile "zip" to create zip artifacts for the release.
- update dlbuilder.sh to specify zip profile explicitly.

Removed the test for Windows NSIS installer because we are not building Windows NSIS installer.


pom.xml
```diff
@@ -148,58 +103,18 @@
...
-      <artifactId>jacob</artifactId>
-      <version>${jacob.version}</version>
-      <classifier>x64</classifier>
-      <type>dll</type>
-      <scope>test</scope>
-    </dependency>
-
-    <dependency>
-      <groupId>quartz</groupId>
+      <groupId>org.quartz-scheduler</groupId>
```
jjangsfdc, line 108: I didn't know Quartz scheduler was used in DL, but okay!

jjangsfdc (approved):
Thank for you cleaning up the POM. Could we please make sure the upgraded modules like quartz, jface and etc are working as expected? Sometimes major version upgrades come with deprecated calls or isn't backward compatible. We can find all the code where they are being used and try to trigger them from the UI or CLI.

ashitsalesforce:
Hi Jae,

Thanks for a quick review. Good point about checking for any undesirable
side-effects of a version upgrade. I will do so.

Ashit

pom.xml
ashitsalesforce: I have verified by running unit tests and integration tests on Mac and manual tests on Mac and Windows.

PR #239, opened 2021-01-08
pom.xml cleanup - consistent formatting, remove unused profiles, and …

ashitsalesforce:
…dependencies

continued pom.xml cleanup - consistent formatting, remove unused profiles, and dependencies

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -31,7 +31,6 @@
     <test.user.restricted>standard@org.com</test.user.restricted>
     <test.password></test.password><!-- use the encrypted password -->
     <test.redirect>false</test.redirect>
-
   </properties>
 
   <repositories>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -468,25 +390,7 @@
           <OSType>mac</OSType>
       </properties>
     </profile>
-    <profile>
```
jjangsfdc, line 471: win32 is old! Frankly, I don't even know if it will work in win 32bits.

pom.xml
```diff
@@ -398,57 +371,6 @@
...
-      </activation>
-      <dependencies>
-        <dependency>
-          <groupId>org.eclipse.swt</groupId>
-          <artifactId>org.eclipse.swt.gtk.linux.x86</artifactId>
-          <version>4.5.1</version>
-        </dependency>
-      </dependencies>
-    </profile>
-    <profile>
```
jjangsfdc, line 417: Removing 32 bit OS profiles - yes. I know we don't officially support Linux but there might be users who are dependent on `linux64`, maybe not. We can remove it or not, your call.

jjangsfdc (approved):

ashitsalesforce:
Thanks Jae. I decided to remove all profiles that we do not maintain/build
with to avoid confusing build failures for the community members in case
they try to build with those profiles.

Ashit

pom.xml
ashitsalesforce: Exactly. Having such obsolete profiles is confusing at best, frustrating if the community members try to use them.

PR #243, opened 2021-01-12
Fixes due to the upgraded Spring framework and log4j

ashitsalesforce:
Fixes include:
- New syntax for log-conf.xml, process-conf.xml, and database-conf.xml in samples folder necessiated by Spring framework and log4j upgrades.

- Fixes in ProcessRunner and ProcessScheduler necessiated by log4j framework upgrade - Controller.initlog() needs to be invoked before instantiating a logger through LogManager.getLogger() call.

- Fixes in Controller.java and version.properties to decouple data loader versioning from the API version used for partner API.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
   <groupId>com.force</groupId>
   <artifactId>dataloader</artifactId>
   <packaging>jar</packaging>
-  <version>51.0.0</version>
+  <version>51.0.1</version>
   <name>Data Loader</name>
   <url>https://github.com/forcedotcom/dataloader</url>
   <organization>
@@ -14,9 +14,9 @@
 
   <properties>
     <short.version>51</short.version>
-    <force.wsc.version>${version}</force.wsc.version>
-    <force.partner.api.version>${version}</force.partner.api.version>
-    <build.year>2020</build.year>
+    <force.wsc.version>${short.version}.0.0</force.wsc.version>
+    <force.partner.api.version>${force.wsc.version}</force.partner.api.version>
+    <build.year>2021</build.year>
     <java.compile.version>11</java.compile.version>
     <minJvmVersion>1.11.0</minJvmVersion>
     <codesign>DONTSIGN</codesign>
--- a/pom.xml
+++ b/pom.xml
@@ -14,7 +14,7 @@
 
   <properties>
     <short.version>51</short.version>
-    <force.wsc.version>${short.version}.0.0</force.wsc.version>
+    <force.wsc.version>51.0.0</force.wsc.version>
     <force.partner.api.version>${force.wsc.version}</force.partner.api.version>
     <build.year>2021</build.year>
     <java.compile.version>11</java.compile.version>
--- a/release/win/samples/conf/database-conf.xml
+++ b/release/win/samples/conf/database-conf.xml
@@ -2,7 +2,8 @@
 <beans>
 <bean id="dbDataSource"
       class="org.apache.commons.dbcp.BasicDataSource"
-      destroy-method="close">
+      destroy-method="close"
+      scope="prototype">
     <property name="driverClassName" value="oracle.jdbc.driver.OracleDriver"/>
     <property name="url" value="jdbc:oracle:thin:@my.server.com:1521:dataBaseName"/>
     <property name="username" value="user"/>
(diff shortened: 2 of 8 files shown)


src/main/java/com/salesforce/dataloader/controller/Controller.java
```diff
@@ -458,23 +467,35 @@ private static String getProductName() {
...
             return;
         }
 
-        File logConfXml = Paths.get(System.getProperty("user.dir"), "configs", LOG_CONF_OVERRIDE).toFile();
-        if (logConfXml.exists()) {
-        	System.setProperty("log4j.configurationFile", logConfXml.getAbsolutePath());
+    	try {
+    		initStaticVariable();
+    	} catch (ControllerInitializationException ex) {
+    		System.out.println("Controller.initLog(): Unable to initialize Controller static vars: " + ex.getMessage());
```
jjangsfdc, line 479: We probably want to pipe this into log4j

src/main/java/com/salesforce/dataloader/controller/Controller.java
```diff
@@ -86,6 +86,7 @@
     private static final String LOG_CONF_OVERRIDE = "log-conf.xml";
 
     private static boolean isLogInitialized = false; // make sure log is initialized only once
+    private static boolean areStaticVarsInitialized = false; // make sure log is initialized only once
```
jjangsfdc, line 89: Minor, usually "isSomething" although the context is plural

pom.xml
```diff
@@ -14,9 +14,9 @@
 
   <properties>
     <short.version>51</short.version>
-    <force.wsc.version>${version}</force.wsc.version>
-    <force.partner.api.version>${version}</force.partner.api.version>
-    <build.year>2020</build.year>
+    <force.wsc.version>${short.version}.0.0</force.wsc.version>
```
jjangsfdc, line 17: How about we create a separate variable like `force.version` then set that to {short.version}.0.0 and then use `force.version` for `force.wsc.version` and `force.partner.api.version` That might be easier to read

release/win/samples/conf/database-conf.xml
```diff
@@ -2,44 +2,46 @@
 <beans>
 <bean id="dbDataSource"
       class="org.apache.commons.dbcp.BasicDataSource"
-      destroy-method="close">
+      destroy-method="close"
+      scope="prototype">
```
jjangsfdc, line 6: Let's make sure these changes do not have any regressions.

src/main/java/com/salesforce/dataloader/controller/Controller.java
```diff
@@ -146,8 +147,11 @@ private String getConfigDirFromArgMap(String[] args) {
                 argNameValuePair.get(CONFIG_DIR_PROP) : null;
     }
 
-    public static void initStaticVariable() throws ControllerInitializationException {
-        Properties versionProps = new Properties();
+    private static synchronized void initStaticVariable() throws ControllerInitializationException {
+    	if (areStaticVarsInitialized) {
```
jjangsfdc, line 151: Indentation

jjangsfdc: Not sure if we need `synchronized` here as DL is a single threaded app

src/main/java/com/salesforce/dataloader/process/ProcessRunner.java
```diff
@@ -98,6 +98,15 @@
 
     private Controller controller;
 
+    static {
+        try {
+            Controller.initLog();
+        } catch (ControllerInitializationException e) {
+            System.out.println("ProcessRunner: Controller not initialized" + e );
```
jjangsfdc, line 105: Same as before, we should pipe it into the log

src/main/java/com/salesforce/dataloader/process/ProcessScheduler.java
```diff
@@ -56,22 +56,30 @@
...
 
     /** Log4J */
-    private static Logger logger = LogManager.getLogger(ProcessScheduler.class);
-
+    private static Logger logger;
+	static {
+        try {
+            Controller.initLog();
+        } catch (ControllerInitializationException e) {
+            System.out.println("ProcessScheduler: Controller not initialized" + e );
```
jjangsfdc, line 76: Same as before

src/main/java/com/salesforce/dataloader/controller/Controller.java
ashitsalesforce: We can't pipe it because log4j is not initialized yet. That's why I am going for System.out.println().

src/main/java/com/salesforce/dataloader/controller/Controller.java
ashitsalesforce: Exactly!

pom.xml
ashitsalesforce: I am realizing that both version and short.version are used for versioning of artifacts, whereas force.wsc.version and force.partner.api.version are used for API version. So best to set these 2 pairs separately by explicitly setting force.wsc.version to a value which may/may not be the same as version.

release/win/samples/conf/database-conf.xml
ashitsalesforce: agree.

src/main/java/com/salesforce/dataloader/controller/Controller.java
ashitsalesforce: Better be (thread) safe :).

src/main/java/com/salesforce/dataloader/process/ProcessRunner.java
ashitsalesforce: Resorting to System.out.println() because the log is not initialized if an exception is thrown.

jjangsfdc (approved):

PR #247, opened 2021-01-16
Derive short version string from full version string instead of setti…

ashitsalesforce:
…ng it in pom.xml

Remove short.version property in pom.xml because it has to be set to the same number as the first number in the version property.


release/mac/dataloader.command
```diff
@@ -1,14 +1,18 @@
 #!/bin/bash
+DATALOADER_VERSION="@@FULL_VERSION@@"
+DATALOADER_SHORT_VERSION=$(echo ${DATALOADER_VERSION} | cut -d'.' -f 1)
+DATALOADER_UBER_JAR_NAME="dataloader-${DATALOADER_VERSION}-uber.jar"
+MIN_JAVA_VERSION=@@MIN_JAVA_VERSION@@
 
 echo ""
 echo "*************************************************************************"
 echo "**            ___  ____ ___ ____   _    ____ ____ ___  ____ ____       **"
```
jjangsfdc, line 9: Minor: we may want to use multiple tabs instead of spaces that way whenever lines are changed it renders neatly
ie) \t

release/mac/dataloader.command
```diff
@@ -18,13 +22,12 @@ echo "**                                                                     **"
 echo "*************************************************************************"
 echo ""
 
-export JAVA_HOME=$(/usr/libexec/java_home -v 11)
```
jjangsfdc, line 21: Good call out, we shouldn't be messing with system's JAVA_HOME.

release/win/dataloader.bat
```diff
@@ -19,23 +26,52 @@ echo **                                                                     **
...
-        echo    eg. ZULU_JAVA_HOME="C:\Program Files\Zulu\zulu-11"
-        PAUSE
-        goto Exit
+:CheckMinJRE
+    echo Data Loader requires Java JRE %MIN_JAVA_VERSION% or later. Checking if it is installed...
+
+    PATH=%PATH%;%JAVA_HOME%\bin\;%ZULU_JAVA_HOME%\bin\;
+
+    java -version 1>nul 2>nul || (
+        goto JavaErrorExit
```
jjangsfdc, line 35: Ahhh, 'goto' statement!

jjangsfdc (approved):
The changes look good. Let's make sure we test the scripts in Mac and Windows.

release/mac/dataloader.command
ashitsalesforce: Yes, I need to figure out a way to address it in IDE or a do a workaround by running a script that replaces tabs with spaces before checking in.

PR #249, opened 2021-01-29
Support for pilot phase of properties to support a feature and PK chunking for bulk query

ashitsalesforce:


ashitsalesforce:
Withdrawing the PR as we are in a strict maintenance mode.

Issue #230, opened 2020-11-20
With Dataloader, incorrect rows displayed in the Success file when an Insert fails with a "Failed to parse date" error


gvpatel:
Hi - Does anyone has solution to this issue - Reference W-4273711 

With Dataloader, incorrect rows displayed in the Success file when an Insert fails with a "Failed to parse date" error

https://trailblazer.salesforce.com/issues_view?id=a1p3A000000mCGXQA2

ashitsalesforce:
Closing the issue as the fix is checked in. It will be available with the next build of data loader, which likely will be v52.

gvpatel:
Thanks ashitsalesforce. Really appreciate it.

hanuma1111:
Hi,
This Issue is not fixed and still facing the same issue with 52 and 54 versions as well.
I have attached the Success and Error files where you can see Error Records are present in Success Table and also, Success Sheet with incorrect data like Customer Class/Cusotmer Sub Class are Required fields for every Account, where as some records with blank or #NA values records are showing in Success sheet without any error.

When you check the Error sheet, error message is incorrect for every record it is showing, where as the data is present.

It's like Error records are swapping into Success Table and also error message also swapped.

Could you please help us to resolve this issue as it is High priority for the Business, because when we generate reports on Transaction History Object for Error Records to trouble shoot, its Miss-Leading.

[LNaccountUpsertError20220301-171824.csv](https://github.com/forcedotcom/dataloader/files/8166495/LNaccountUpsertError20220301-171824.csv)
[LNaccountUpsertSuccess20220301-171824.csv](https://github.com/forcedotcom/dataloader/files/8166496/LNaccountUpsertSuccess20220301-171824.csv)

hanuma1111:
I have created another Issue 436 for the same issue as it is closed, if possible, could you please take a look and provide your inputs which will help to resolve the issue.

https://github.com/forcedotcom/dataloader/issues/436

PR #253, opened 2021-03-06
Fix for issue #230 - keep track of rows that had format conversion error when preparing for a bulk operation.

ashitsalesforce:


src/main/java/com/salesforce/dataloader/action/visitor/BulkLoadVisitor.java
```diff
@@ -328,11 +327,28 @@ private void processResults(final DataReader dataReader, final BatchInfo batch,
...
+        // from the DataReader
+        while (rowCount < clientBatchInfo.numRows) {
+            boolean conversionSuccessOfRow = isRowConversionSuccessful(dataReaderRowNumBase + rowCount + additionalRowsToReadFromDataReader);
+            if (conversionSuccessOfRow) {
+                rowCount++;
+            } else {
+                additionalRowsToReadFromDataReader++;
+            }
+        }
+        totalRowsToReadFromDataReader += additionalRowsToReadFromDataReader;
```
jjangsfdc, line 346: Might be easier to follow if it was initialized here
```suggestion
        final int totalRowsToReadFromDataReader = clientBatchInfo.numRows + additionalRowsToReadFromDataReader;
```

src/main/java/com/salesforce/dataloader/action/visitor/BulkQueryVisitor.java
```diff
@@ -100,9 +104,33 @@ private void writeExtractionForBatch(BatchInfo batch) throws AsyncApiException,
 
         for (final String resultId : results.getResult()) {
             if (getProgressMonitor().isCanceled()) return;
+            boolean bufferResults = getController().getConfig().getBoolean(Config.BUFFER_UNPROCESSED_BULK_QUERY_RESULTS);
```
jjangsfdc, line 107: Does this work?
```suggestion
            boolean bufferResults = getConfig().getBoolean(Config.BUFFER_UNPROCESSED_BULK_QUERY_RESULTS);
```

jjangsfdc: Also, do you need to call it for every `for` iteration?

src/main/java/com/salesforce/dataloader/action/visitor/BulkLoadVisitor.java
```diff
@@ -81,7 +81,6 @@
     private static final String ERROR_RESULT_COL = "Error";
     private static final String ID_RESULT_COL = "Id";
     private static final String CREATED_RESULT_COL = "Created";
-    private static final String SKIP_BATCH_ID = "SKIP";
```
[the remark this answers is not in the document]
ashitsalesforce: The new mechanism to keep the track of each row to be processed on server side eliminates the need to have batches created in DL (not created by making a call to the server) with "SKIP" ID that were used to track the rows that failed local processing and format conversion. Consequently, the related constant is not used by the code anymore.

src/main/java/com/salesforce/dataloader/action/visitor/BulkLoadVisitor.java
ashitsalesforce: good point. I will make that change.

src/main/java/com/salesforce/dataloader/action/visitor/BulkQueryVisitor.java
ashitsalesforce: another good catch (2 of them). I will make the change.

Issue #270, opened 2021-05-27
Data Loader installation fails with “Invalid path”


isherwood:
I've downloaded and extracted Data Loader for Windows (v52) from my org. OpenJDK Azul (Java 15) is installed. Upon running install.bat I'm prompted for an installation path. I enter Code\Salesforce\DataLoader, for example. File copy initiates but fails immediately with this message:

> Copying files from 'C:\Users\my.name\Downloads\dataloader_win' to 'C:\Users\my.name\Code\Salesforce\DataLoader\v52.0.0' ...
> Invalid path
> 0 File(s) copied
> The system cannot find the file specified. Your Data Loader v52.0.0 is created in 'C:\Users\my.name\Code\Salesforce\DataLoader\v52.0.0'

The process then asks about creating shortcuts and exits.

The directory exists. If I tab through entry of a full path all parent directories are found. Still it fails. What's the trouble?

Update: I replaced install.bat with the latest version from the repo. No change.

ashitsalesforce:
Hi isherwood, are you seeing the same issue with an earlier Data Loader version such as v51.0.1? Also, did replacing install.bat with the latest repo version work? Likely not but want to confirm.

Thanks.

isherwood:
I mentioned replacing install.bat in my post. That didn't help. However, v51.0.1 _did_ install successfully.

I also tried v52.0.0 downloaded directly from the repo and had the same failure.

ashitsalesforce:
Hi isherwood,

I was able to reproduce the issue and have a potential fix that you may want to try. I am attaching the install.bat file containing the potential fix. It will be great if you can try and let me know if it addresses the issue. I have renamed it install.txt. Change the extension to .bat after downloading it to run it.
[install.txt](https://github.com/forcedotcom/dataloader/files/6557726/install.txt)


Thanks.

ashitsalesforce:
Hi @isherwood , I have been looking into why xcopy command used by installer.bat fails to copy with the error "0 files copied" on some Windows OS (Windows Server 2019 in my experience) and works on some other OS (Windows 10 in my experience). Perhaps the root-cause is OS version related. It will help if you can specify the OS version you are using.

Thanks.

Issue #269, opened 2021-05-25
Typo in the Dataloader Login page screen message


rsikhakolli:
We have received a Trailhead Help case where the user reported a typo in the Dataloader login page. Please check the below screenshot for the typo:
<img width="698" alt="Typo" src="https://user-images.githubusercontent.com/61190344/119478543-84153780-bd6d-11eb-9eff-e87543eade12.png">

The message in Step1 has a typo "Log in to the Salesforce org that you?re importing to or exporting from." and it should be "Log in to the Salesforce org that you're importing to or exporting from."
Can someone please check and fix the typo? Thanks in advance! 

Note: I have installed DataLoader 51.0.1 in Mac Book Air and seeing the same typo on the DataLoader login screen.

ashitsalesforce:
Fixed the typo. Fix will be available in the next release.

Thanks.

PR #271, opened 2021-05-28
fix for issues #270 and #269

ashitsalesforce:
fix for issues #270 and #269
- Fix for the issue #270 in install.bat by removing the trailing \ character from the value of %~dp0.

- Fix for an issue that caused "The system cannot find the batch label specified - JavaErrorExit" message when running dataloader.bat without installing a JRE on Windows.

- Fix for issue #269


jjangsfdc (approved):
LGTM

shireennagdive (approved):

Issue #272, opened 2021-06-15
[Feature Request] Further improvements regarding Object Search mechanism


pgumiela:
This FR is a continuation of #261.

It would be great to consider the following improvements:
- the ability to search both by Label or API Name (currently it works only for API Name)
- narrow down search results immediately, without the need to press any keys
- add support for wildcard characters (such as \*) to make the search easier (like Acc*)

ashitsalesforce:
Added the following improvements:
- the ability to search both by Label or API Name (currently it works only for API Name)
- narrow down search results immediately, without the need to press any keys

These will be available in the next release of the data loader. Did not add wildcard characters because the search looks for the pattern match rather than exact match.

PR #273, opened 2021-06-19
fix for #272 improvements regarding Object Search

ashitsalesforce:
Changes to address the following:

- the ability to search both by Label or API Name (currently it works only for API Name)

- narrow down search results immediately, without the need to press any keys


reubencornel (approved):

src/main/java/com/salesforce/dataloader/ui/entitySelection/EntityFilter.java
```diff
@@ -70,19 +72,20 @@ public boolean select(Viewer arg0, Object arg1, Object arg2) {
...
     }
     
-    private boolean filterBySearchText(String entityName) {
+    private boolean filterBySearchText(String entityName, String entityLabel) {
         String searchText = this.searchText.getText();
         if (searchText != null && !searchText.isEmpty() && entityName != null) {
             searchText = searchText.toLowerCase();
-            if (entityName.toLowerCase().contains(searchText)) {
+            if (entityName.toLowerCase().contains(searchText) 
+                    || entityLabel.toLowerCase().contains(searchText)) {
```
lzemskov, line 88: Do we also need a null check for entityLabel?

shireennagdive (approved):

ashitsalesforce:
good point. I will make the change.

src/main/java/com/salesforce/dataloader/ui/entitySelection/EntityFilter.java
ashitsalesforce: Done. please review.

reubencornel: There is no need for a null check here, it does not hurt, but its not required. This value comes from the describe result so it cannot be null.

Issue #280, opened 2021-07-08
Dates automatically go back one day after the update


HUMENTH:
When I update the data, the dates automatically go back one day after the update, it only happens with the field with the date, not the date/time field.

> **Device: Desktop
> OS: Windows 10
> Date format: dd/MM/yyyy
> Dataloader ver.: 52.0.0
> Dataloader Time Zone: Asia/Calcutta**
![Dataloader Settings](https://user-images.githubusercontent.com/4609678/124881549-19524f80-dfed-11eb-8977-1ed78fc3a8c6.jpg)

ashitsalesforce:
Hi @HUMENTH , provide a sample CSV containing the dates that you are updating to help troubleshoot and attach a screenshot of the updated record with incorrect date to aid in troubleshooting the issue.

Also, what is the timezone of the user in the Salesforce whose credentials you are using? You can see this information by clicking on "View profile" from the dropdown when you hover on the user icon in the top-right corner of the screen after logging into your Salesforce org using that user's credentials.

Thanks.

HUMENTH:
Hello @ashitsalesforce, 
Please review attached CSV and Screenshots.

I'm using the Salesforce administrator User.

![Salesforce screenshot with wrong dates](https://user-images.githubusercontent.com/4609678/125036280-72cf8280-e0b0-11eb-8b09-5c9e3fd57c76.jpg)
![TimeZone and Locale of the SFDC User](https://user-images.githubusercontent.com/4609678/125036287-74994600-e0b0-11eb-88ed-0d9aea6bb959.jpg)

[CSV](https://drive.google.com/open?id=1HDjjlwit5P2Yf-V77lpSpJ-RNvR7IJDO&authuser=CRM%40respoproducts.com&usp=drive_fs)

ashitsalesforce:
Fix checked in. It will be available in v53.

PR #285, opened 2021-07-16
Fix for issue #280

ashitsalesforce:
Fix for #280 (Dates automatically go back one day after the update) by mapping xsd:date to DateOnlyCalendar which adjusts for timezones ahead of GMT.

### DIFF
--- a/src/main/java/com/salesforce/dataloader/controller/Controller.java
+++ b/src/main/java/com/salesforce/dataloader/controller/Controller.java
@@ -73,6 +73,7 @@ import java.util.Date;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Properties;
+import java.util.Set;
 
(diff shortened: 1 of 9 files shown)


src/main/java/com/salesforce/dataloader/controller/Controller.java
```diff
@@ -108,6 +109,8 @@
 
     private static AppUtil.OSType OS_TYPE;
     private static boolean reuseClientConnection = true;
+
```
shireennagdive, line 112: nit : can remove empty line

ashitsalesforce: I had the empty line for readability. To show that the newly added class-level field is for a distinct use-case than the previous ones.

ashitsalesforce:
Withdrawing PR. Found a better fix.

PR #287, opened 2021-07-17
Fix for issue #280

ashitsalesforce:
Fix for #280 (Dates automatically go back one day after the update) by mapping xsd:date to Date instead of Calendar and using GMT as the locale.


ashitsalesforce:
Merging the PR based on feedback from @ktakazawa.

PR #289, opened 2021-07-26
upfront checks in the batch mode. Minor improvements to installation scripts.

ashitsalesforce:
- Upfront checks in the batch mode to make sure that the required parameters are specified either in the process DynaBean or in the config.properties file.

- Minor improvements in the installation scripts to not deploy .ico, META-INF, and other unused files in the installation folder.

### DIFF
--- a/src/main/java/com/salesforce/dataloader/process/ProcessRunner.java
+++ b/src/main/java/com/salesforce/dataloader/process/ProcessRunner.java
@@ -78,8 +78,6 @@ import java.util.Calendar;
 import java.util.HashMap;
 import java.util.Map;
 
-import javax.xml.parsers.FactoryConfigurationError;
-
 public class ProcessRunner implements InitializingBean, Job, Runnable {
 
     /**
@@ -89,14 +87,23 @@ public class ProcessRunner implements InitializingBean, Job, Runnable {
 
     //logger
     private static Logger logger;
-
-    // Name of the current engine runner.  Improves readability of the log output
-    String name;
+    
+    private String name; // name of the loaded process DynaBean
 
     // config override parameters
     private final Map<String, String> configOverrideMap = new HashMap<String, String>();
 
     private Controller controller;
+    
+    private static final String PROP_NAME_ARRAY[] = {
+            Config.OPERATION,
+            Config.ENDPOINT,
+            Config.USERNAME,
+            Config.PASSWORD,
+            Config.DAO_TYPE,
+            Config.DAO_NAME,
+            Config.ENTITY,
+    };
     
     static {
 
@@ -131,6 +138,9 @@ public class ProcessRunner implements InitializingBean, Job, Runnable {
             // load parameter overrides (from command line or caller context)
             logger.info(Messages.getString("Process.loadingParameters")); //$NON-NLS-1$
             config.loadParameterOverrides(getConfigOverrideMap());
+            
+            // Make sure that the required properties are specified.
+            validateConfigProperties(config);
 
             // create files for status output unless it's an extract and status output is disabled
             if (!config.getOperationInfo().isExtraction() || config.getBoolean(Config.ENABLE_EXTRACT_STATUS_OUTPUT)) {
@@ -243,7 +253,11 @@ public class ProcessRunner implements InitializingBean, Job, Runnable {
     }
 
     private static void topLevelError(String message, Throwable err) {
-        logger.fatal(message, err);
+        if (logger == null) {
+            System.err.println(message);
+        } else {
+            logger.fatal(message, err);
+        }
         System.exit(-1);
     }
     
@@ -283,7 +297,7 @@ public class ProcessRunner implements InitializingBean, Job, Runnable {
      * @return instance of ProcessRunner
      * @throws ProcessInitializationException
      */
-    public static ProcessRunner getInstance(Map<String, String> argMap) throws ProcessInitializationException {
+    public static synchronized ProcessRunner getInstance(Map<String, String> argMap) throws ProcessInitializationException {
         ProcessRunner runner;
         Controller.setConfigDir(argMap);
         try {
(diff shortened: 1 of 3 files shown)


ashitsalesforce:
The PR also contains the fix for issue #288.

src/main/java/com/salesforce/dataloader/process/ProcessRunner.java
```diff
@@ -294,17 +308,37 @@ public static ProcessRunner getInstance(Map<String, String> argMap) throws Proce
...
+            Config config = controller.getConfig();
+            // load parameter overrides (from command line or caller context)
+            logger.info(Messages.getString("Process.loadingParameters")); //$NON-NLS-1$
+            config.loadParameterOverrides(argMap);
+            String processName = config.getString(PROCESS_NAME);
+            logger.debug("process name is " + processName);
+            if (processName == null || processName.isEmpty()) {
+                logger.info(PROCESS_NAME + "is not set in the command line or config.properties file.");
+                // operation and other process params are specified through properties
+                if (!validateConfigProperties(config)) {
```
lzemskov, line 325: Is this condition ever hit?
Looks like the outcome of the method validateConfigProperties() will be either 'TRUE' or ProcessInitializationException.

src/main/java/com/salesforce/dataloader/process/ProcessRunner.java
```diff
@@ -294,17 +308,37 @@ public static ProcessRunner getInstance(Map<String, String> argMap) throws Proce
...
+                    runner.setName(config.getString(Config.OPERATION));
+                    runner.setConfigOverrideMap(argMap);
+                }
+            } else {
+                // process DynaBean name specified.
+                runner = ProcessConfig.getProcessInstance(processName);
+                runner.getConfigOverrideMap().putAll(argMap);
+            }
+        } catch (Exception e) {
+            throw new RuntimeException(e);
```
lzemskov, line 339: Current method declares that it throws ProcessInitializationException, but looks like it will no longer be happening because we swallow all exceptions inside this clause and throw RuntimeException up the chain. Should this be 2 clauses? one to catch and re-throw ProcessInitializationException and the other one to throw new RuntimeException()

Also looking through this try/catch what else do we expect to throw an 'Exception' which we need to catch, why do we need this clause?

src/main/java/com/salesforce/dataloader/process/ProcessRunner.java
```diff
@@ -294,17 +308,37 @@ public static ProcessRunner getInstance(Map<String, String> argMap) throws Proce
...
+                    logger.fatal(Messages.getFormattedString("Process.missingRequiredArg", PROCESS_NAME));
+                    throw new ParameterLoadException(Messages.getFormattedString("Process.missingRequiredArg", PROCESS_NAME));
+                } else {
+                    runner = new ProcessRunner();
+                    runner.setName(config.getString(Config.OPERATION));
+                    runner.setConfigOverrideMap(argMap);
+                }
+            } else {
+                // process DynaBean name specified.
+                runner = ProcessConfig.getProcessInstance(processName);
```
lzemskov, line 335: do we need a null check for 'runner' or it will never be null?

src/main/java/com/salesforce/dataloader/process/ProcessRunner.java
```diff
@@ -243,7 +253,11 @@ private static boolean validateCmdLineArgs (String[] args) {
     }
 
     private static void topLevelError(String message, Throwable err) {
-        logger.fatal(message, err);
+        if (logger == null) {
+            System.err.println(message);
```
lzemskov, line 257: why System.err instead of System.out are there 2 different destinations? Looks like all other print statements go to System.our even for error conditions.

src/main/java/com/salesforce/dataloader/process/ProcessRunner.java
```diff
@@ -131,6 +138,9 @@ public synchronized void run(ILoaderProgress monitor) {
             // load parameter overrides (from command line or caller context)
             logger.info(Messages.getString("Process.loadingParameters")); //$NON-NLS-1$
             config.loadParameterOverrides(getConfigOverrideMap());
+            
+            // Make sure that the required properties are specified.
+            validateConfigProperties(config);
```
lzemskov, line 143: This method throws ProcessInitializationException which has useful info, so it might be better to handle it similar to what we do for ApiFault below, where we get the exception message from that Exception otherwise this info will be lost.

src/main/java/com/salesforce/dataloader/process/ProcessRunner.java
ashitsalesforce: Good point. How about changing validateConfigProperties to not return a boolean and change the code from
`if (!validateConfigProperties(config)) {`

to

`validateConfigProperties(config);`

src/main/java/com/salesforce/dataloader/process/ProcessRunner.java
ashitsalesforce: Good catch! Removing try/catch block makes the most sense.

src/main/java/com/salesforce/dataloader/process/ProcessRunner.java
ashitsalesforce: It will never be null. A failure to instantiate the runner would result in an exception thrown.

src/main/java/com/salesforce/dataloader/process/ProcessRunner.java
ashitsalesforce: I would switch others that are using System.out to print error message to System.err. However, that would be a different fix.

src/main/java/com/salesforce/dataloader/process/ProcessRunner.java
ashitsalesforce: I do not follow this point. The catch block on line 184 picks up the exception message when creating an instance of RuntimeException:
` catch (Exception e) {
            throw new RuntimeException(e);
        }`

src/main/java/com/salesforce/dataloader/process/ProcessRunner.java
ashitsalesforce: updated the PR with the change.

src/main/java/com/salesforce/dataloader/process/ProcessRunner.java
ashitsalesforce: updated the code.

src/main/java/com/salesforce/dataloader/process/ProcessRunner.java
ashitsalesforce: updated to consistently use stderr when logging an error message prior to initializing Logger.

src/main/java/com/salesforce/dataloader/process/ProcessRunner.java
ashitsalesforce: I am proceeding to merge the PR.  I can fix issues, in case there are any that come up post-merge.

PR #292, opened 2021-08-06
Support for Bulk V2 query - pilot

ashitsalesforce:
This change contains the usage of bulk v2 query if enabled with associated UI changes to select the option as well as tests to check the code.


src/main/java/com/salesforce/dataloader/action/visitor/BulkV2Connection.java
```diff
@@ -0,0 +1,364 @@
...
+import com.sforce.ws.MessageHandler;
+import com.sforce.ws.MessageHandlerWithHeaders;
+import com.sforce.ws.bind.CalendarCodec;
+import com.sforce.ws.parser.PullParserException;
+import com.sforce.ws.parser.XmlInputStream;
+import com.sforce.ws.parser.XmlOutputStream;
+import com.sforce.ws.transport.Transport;
+import com.sforce.ws.util.FileUtil;
+
+public class BulkV2Connection extends BulkConnection {
```
mgrumbach, line 74: What is the motivation for using inheritance here? Can you use composition instead?

src/main/java/com/salesforce/dataloader/action/visitor/BulkV2Connection.java
```diff
@@ -0,0 +1,364 @@
...
+    private static final JsonFactory JSON_FACTORY = new JsonFactory(new ObjectMapper());
+    private static final String URI_STEM_QUERY = "query/";
+    private static final String AUTH_HEADER = "Authorization";
+    private static final String AUTH_HEADER_VALUE_PREFIX = "Bearer ";
+    
+    private String authHeaderValue = "";
+    private String queryLocator = "";
+    private String queryResultNumberOfRecords = "";
```
mgrumbach, line 82: Is this used anywhere?

src/main/java/com/salesforce/dataloader/action/visitor/BulkV2Connection.java
```diff
@@ -0,0 +1,364 @@
...
+
+    public BulkV2Connection(ConnectorConfig connectorConfig) throws AsyncApiException {
+        super(connectorConfig);
+    }
+    
+    public JobInfo createJob(JobInfo job) throws AsyncApiException {
+        return createOrUpdateJob(job, job.getContentType() == null ? ContentType.JSON
```
mgrumbach, line 89: I believe V2 Ingest only supports CSV jobs. Maybe you should simplify the new code here by removing any code that handles other job types.

src/main/java/com/salesforce/dataloader/action/visitor/BulkV2Connection.java
```diff
@@ -0,0 +1,364 @@
...
+        HashMap<String, String> headers = getHeaders(JSON_CONTENT_TYPE);
+
+        // Bulk v2 uses OAuth 2 access token for Auth
+
+        try {
+            Transport transport = getConfig().createTransport();
+            OutputStream out;
+            if (contentType == ContentType.XML || contentType == ContentType.ZIP_XML) {
+                out = transport.connect(getConfig().getRestEndpoint() + URI_STEM_QUERY, headers);
+                XmlOutputStream xout = new AsyncXmlOutputStream(out, true);
```
mgrumbach, line 108: xout is created then immediately closed.

src/main/java/com/salesforce/dataloader/action/visitor/BulkV2Connection.java
```diff
@@ -0,0 +1,364 @@
...
+            : job.getContentType());
+    }
+    
+    private JobInfo createOrUpdateJob(JobInfo job, ContentType contentType) throws AsyncApiException {
+        String soqlStr = job.getObject();
+        
+        // hardcoded header for OAuth2 auth bearer token
+        // TODO: figure a way to get OAuth2 auth bearer token from session id
+        this.authHeaderValue = AUTH_HEADER_VALUE_PREFIX + getConfig().getSessionId();
+        HashMap<String, String> headers = getHeaders(JSON_CONTENT_TYPE);
```
mgrumbach, line 99: This header setting assumes JSON. So it would contradict the XML case below.

src/main/java/com/salesforce/dataloader/action/visitor/BulkV2Connection.java
```diff
@@ -0,0 +1,364 @@
...
+                    type = ContentType.XML;
+                } else if (contentTypeHeader.contains(JSON_CONTENT_TYPE)) {
+                    type = ContentType.JSON;
+                }
+            }
+            parseAndThrowException(in, type);
+        }
+
+        this.queryLocator = connection.getHeaderField("Sforce-Locator");
+        this.queryResultNumberOfRecords = connection.getHeaderField("Sforce-NumberOfRecords");
```
mgrumbach, line 307: This method seems fairly generic and usable for different GET requests. But these last two lines are specific to GET requests for query results. Can these lines be moved somewhere else?

src/main/java/com/salesforce/dataloader/action/visitor/BulkV2Connection.java
```diff
@@ -0,0 +1,364 @@
...
+
+    public JobInfo closeJob(String jobId) throws AsyncApiException {
+        
+        JobInfo job = new JobInfo();
+        job.setId(jobId);
+        job.setState(JobStateEnum.Closed);
+        return job;
+        /*
+        return updateJob(job);
+        */
```
mgrumbach, line 323: There is no "close" operation for bulk v2 jobs. So maybe this method should throw UnsupportedOperationException.

src/main/java/com/salesforce/dataloader/action/visitor/BulkV2Connection.java
```diff
@@ -0,0 +1,364 @@
...
+    }
```
mgrumbach, line 324: But Bulk V2 does support deleting jobs. So maybe a deleteJob() method should be added.

src/main/java/com/salesforce/dataloader/action/visitor/BulkV2Connection.java
```diff
@@ -0,0 +1,364 @@
...
+        mapper.writeValue(generator, value);
+    }
+    
+    static <T> T deserializeJsonToObject (InputStream in, Class<T> tmpClass) throws IOException {
+        ObjectMapper mapper = new ObjectMapper();
+        mapper.configure(DeserializationFeature.ACCEPT_SINGLE_VALUE_AS_ARRAY, true);
+        // By default, ObjectMapper generates Calendar instances with UTC TimeZone.
+        // Here, override that to "GMT" to better match the behavior of the WSC XML parser.
+        mapper.setTimeZone(TimeZone.getTimeZone("GMT"));
+        InputStreamReader isReader = new InputStreamReader(in);
```
mgrumbach, line 178: Why does the stream need to be materialized into a string?

src/main/java/com/salesforce/dataloader/action/visitor/BulkApiVisitorUtil.java
```diff
@@ -154,6 +169,10 @@ void createJob(Config cfg) throws AsyncApiException {
                                       "chunkSize=" + this.queryChunkSize + startRowParam);
             }
         }
+        if (isBulkV2QueryJob()) {
+            job.setObject(this.config.getString(Config.EXTRACT_SOQL));
```
mgrumbach, line 173: I am guessing that "EXTRACT_SOQL" is the soql statement. Is that the right config param to assign to "Object"?

ashitsalesforce: I agree that this is not a "clean" fix. Unfortunately, JobInfo does not have a field for SoQL because BulkConnection has an explicit method, createBatch, that accepts SoQL as an argument. There in no notion of a batch in Bulk V2. So, the best way to address this side-effect seems to be to extend JobInfo, perhaps as BulkV2JobInfo through inheritance? containment would require a lot of boilerplate code for all public methods of JobInfo.

src/main/java/com/salesforce/dataloader/action/visitor/BulkV2Connection.java
ashitsalesforce: Agreed. I will skip closing a job for bulk v2 query and remove the method from BulkV2Connection.java.

src/main/java/com/salesforce/dataloader/action/visitor/BulkV2Connection.java
ashitsalesforce: Agreed. I have split the method and its usage as follows:

    
```
    private InputStream doGetQueryResultStream(URL resultsURL) throws IOException, AsyncApiException {
        HttpURLConnection httpConnection = openHttpConnection(resultsURL);
        InputStream is = doHttpGet(httpConnection, resultsURL);
        this.queryLocator = httpConnection.getHeaderField("Sforce-Locator");
        this.numberOfRecordsInQueryResult = Integer.valueOf(httpConnection.getHeaderField("Sforce-NumberOfRecords"));
        return is;
    }
    
    private InputStream doGetJobStatusStream(URL jobStatusURL) throws IOException, AsyncApiException {
        HttpURLConnection httpConnection = openHttpConnection(jobStatusURL);
        return doHttpGet(httpConnection, jobStatusURL);
    }
    
    private HttpURLConnection openHttpConnection(URL url) throws IOException {
        HttpURLConnection connection = getConfig().createConnection(url, null);
        SSLContext sslContext = getConfig().getSslContext();
        if (sslContext != null && connection instanceof HttpsURLConnection) {
            ((HttpsURLConnection)connection).setSSLSocketFactory(sslContext.getSocketFactory());
        }
        connection.setRequestProperty(AUTH_HEADER, this.authHeaderValue);
        return connection;
    }
    
    private InputStream doHttpGet(HttpURLConnection connection, URL url) throws IOException, AsyncApiException {
        boolean success = true;
        InputStream in;
        try {
            ...
            ...
```

src/main/java/com/salesforce/dataloader/action/visitor/BulkApiVisitorUtil.java
ashitsalesforce: Will await extending JobInfo with a member variable to store the SoQL string and use it instead of setting its Object variable to store the query string to be passed during Bulk V2 query job creation.

src/main/java/com/salesforce/dataloader/action/visitor/BulkV2Connection.java
ashitsalesforce: Good point. I have removed closeJob() from BulkV2Connection.

src/main/java/com/salesforce/dataloader/action/visitor/BulkV2Connection.java
ashitsalesforce: Data Loader does not delete bulk jobs. So, I will not add it to BulkV2Connection.java for now.

src/main/java/com/salesforce/dataloader/action/visitor/BulkV2Connection.java
ashitsalesforce: Yes, I have refactored doHttpGet() to be able to get the handle to HttpURLConnection object required to get the header fields and moved the 2 lines to the calling method.

ashitsalesforce:
Going ahead with the merge after fixing the issues based on review feedback. The feature is not enabled as Data Loader is in maintenance mode. However, checking it in as the code provides a future possible enhancement to WSC (BulkV2Connection.java) and its use on the client side. 

To enable the feature, take the following steps:
- Remove the following lines from com.salesforce.dataloader.ui.AdvancedSettingsDialog.java:
```
        // hide the label to enable/disable Bulk V2 query
        labelUseBulkV2Query.setVisible(false);
```
and
```
        
        // hide the checkbox to enable/disable Bulk V2 query
        buttonUseBulkV2Query.setVisible(false);
```

- Remove the following lines from com.salesforce.dataloader.config.Config.skipSave() (Config.java)
```
        // do not save a value for enabling Bulk V2 query
        this.properties.remove(ENABLE_BULK_V2_QUERY);

```

Issue #410, opened 2022-01-14
SSO Login fails with dataloader opened through the .bat


gulinski-krystian:
Versions: v52+
Platform: Windows
SSO provider: PingIdentity
java: zulu11

So the issue appears when you open dataloader through the batch file, which most users who installed the dataloader will use.

Steps:
1. Click on any action
2. Prompt the OAuth login (any env)
3. Webview pops up, you follow the flow and use the SSO
4. After you finish logging in the last redirect doesn't work properly and it gets stuck at the screen attached.
5. If you close the window and attempt to log in again it will work, so only the first login gets stuck

Important note: Issue doesn't happen if you open dataloader directly through the .jar file

---

Tried opening from cmd and it seems like the option added in the batch file is causing this.
`-D"org.eclipse.swt.browser.IEVersion=12001"`
If i remove the option it's working as expected.

![image](https://user-images.githubusercontent.com/73952595/149515403-e160bce2-cdd7-4d20-84ef-b7cc7157c933.png)

ashitsalesforce:
Hi @gulinski-krystian , the java system property org.eclipse.swt.browser.IEVersion is set to 12001 so that SWT uses Edge HTML engine in its Browser widget, which is used by Data Loader when OAuth login is selected. 

If this system property is not set, the default value used by SWT is IE 11. Salesforce has stopped supporting IE 11 - see [announcement](https://help.salesforce.com/s/articleView?id=release-notes.rn_supported_browsers_ie11.htm&language=en_US&r=https%3A%2F%2Fwww.google.com%2F&release=230&type=5). Not setting this system property had resulted in issues #242 and #196 when OAuth provider is Salesforce. Given that you had to switch to the default value for your OAuth provider, PingIdentity, wondering if their OAuth client-side javascript does not work on Edge browser.

ashitsalesforce:
Hi @gulinski-krystian , -skipieversion is added to dataloader.bat to skip setting of org.eclipse.swt.browser.IEVersion. This should address your issue.

ashitsalesforce:
The checked in fix will let an explicit setting of org.eclipse.swt.browser.IEVersion system property override the default value of 12001. Set JAVA_TOOL_OPTIONS environment variable to the desired value. For example, set JAVA_TOOL_OPTIONS to "-Dorg.eclipse.swt.browser.IEVersion=11001" if IE browser instead of Edge browser is desired.

gulinski-krystian:
[...]

PR #411, opened 2022-01-14
Change IE compat value in the windows batch file to a valid value

gulinski-krystian:
Fixes: https://github.com/forcedotcom/dataloader/issues/410

According to [this part of the docs](https://www.eclipse.org/swt/faq.php#:~:text=Additionally%2C%20the%20Browser%27s,mode%20by%20default.) the value set for this option is one of the values listed [here](https://docs.microsoft.com/en-us/previous-versions/windows/internet-explorer/ie-developer/general-info/ee330730(v=vs.85)?redirectedfrom=MSDN#browser_emulation)

I've found multiple references online to 12001 being the correct value for ie11 edge mode, but no reliable source that would actually confirm this

### DIFF
--- a/release/win/dataloader.bat
+++ b/release/win/dataloader.bat
@@ -54,7 +54,7 @@ IF NOT "%DATALOADER_JAVA_HOME%" == "" (
 
 :Run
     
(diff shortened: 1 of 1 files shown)


ashitsalesforce:
closing the PR as a fix is checked in.

Issue #494, opened 2022-06-22
java.lang.ClassCastException: class com.salesforce.dataloader.ui.extraction.ExtractionSOQLPage cannot be cast to class com.salesforce.dataloader.ui.extraction.ExtractionFinishPage


matinlotfali:
I am getting the following error when trying to Export.

```
java.lang.ClassCastException: class com.salesforce.dataloader.ui.extraction.ExtractionSOQLPage cannot be cast to class com.salesforce.dataloader.ui.extraction.ExtractionFinishPage (com.salesforce.dataloader.ui.extraction.ExtractionSOQLPage and com.salesforce.dataloader.ui.extraction.ExtractionFinishPage are in unnamed module of loader 'app')
	at com.salesforce.dataloader.ui.extraction.ExtractionWizard.getFinishPage(ExtractionWizard.java:79)
	at com.salesforce.dataloader.ui.extraction.ExtractionWizard.performFinish(ExtractionWizard.java:128)
	at com.salesforce.dataloader.ui.LoaderWizardDialog.finishPressed(LoaderWizardDialog.java:650)
	at com.salesforce.dataloader.ui.LoaderWizardDialog.buttonPressed(LoaderWizardDialog.java:314)
	at com.salesforce.dataloader.ui.LoaderWizardDialog$5.widgetSelected(LoaderWizardDialog.java:1259)
	at org.eclipse.swt.widgets.TypedListener.handleEvent(TypedListener.java:252)
	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:89)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:5794)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1529)
	at org.eclipse.swt.widgets.Display.runDeferredEvents(Display.java:5025)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:4477)
	at org.eclipse.jface.window.Window.runEventLoop(Window.java:823)
	at org.eclipse.jface.window.Window.open(Window.java:799)
	at com.salesforce.dataloader.ui.uiActions.OperationUIAction.run(OperationUIAction.java:54)
	at com.salesforce.dataloader.ui.LoaderWindow$2.widgetSelected(LoaderWindow.java:257)
	at org.eclipse.swt.widgets.TypedListener.handleEvent(TypedListener.java:252)
[...]

PR #495, opened 2022-06-23
Fix cast in ExtractionWizard.java

matinlotfali:
This fixes issue #494


ashitsalesforce:
closing this PR because while it fixes this issue, it creates another similar issue if `Generate status files for exports` is enabled in Settings.

PR #552, opened 2022-08-17
support for "keep account team" and api version in config.properties

ashitsalesforce:
- Added process.keepAccountTeam property for SOAP Partner operations per ask at https://ideas.salesforce.com/s/idea/a0B8W00000GdYyxUAF/dataloader-account-owner-change-keep-account-team-data-loader-45-and-above

- Added "salesforce.api.version" property, which can be set in config.properties to specify an API version that is different than WSC version.

### DIFF
--- a/src/main/java/com/salesforce/dataloader/client/ClientBase.java
+++ b/src/main/java/com/salesforce/dataloader/client/ClientBase.java
@@ -81,7 +81,7 @@ public abstract class ClientBase<ClientType> {
         this.controller = controller;
         this.config = controller.getConfig();
         this.logger = logger;
-        String apiVersionStr = config.getString(Config.CLI_OPTION_API_VERSION);
+        String apiVersionStr = config.getString(Config.API_VERSION_PROP);
         if (apiVersionStr != null && !apiVersionStr.isEmpty()) {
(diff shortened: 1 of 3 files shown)


mrichbb:
@ashitsalesforce we have some business users that want to use this property, but it appears that it is not exposed via the Settings UI. We tried setting the value in the config.properties file, but it appears to be overwritten to the default (false) before the operation is attempted. 

What is the suggested way for this to be used through the UI?

ashitsalesforce:
Hi @mrichbb, 

You are right that `process.keepAccountTeam` is not exposed via the Settings UI yet. Given the ask, we will consider this as a feature enhancement request and file it so that it gets prioritized for a future release.

As you it getting overridden, our internal tests and its use by other customers do not show this issue. So, we need your help in troubleshooting it further. Are you launching data loader first and then setting the property `process.keepAccountTeam=true` in config.properties file or are you setting the property first and then launching data loader?

mrichbb:
@ashitsalesforce so the trick is to set the property before launching the application? It appears that it is re-writing the properties file multiple times throughout the steps needed to update records, but only keeps the setting if I had set it before opening.

PR #563, opened 2022-10-14
fix for W-11901215 - Unable to connect to DB using ojdbc11.jar from D…

ashitsalesforce:
…ataLoader

Fix for the issue of Unable to connect to DB using ojdbc11.jar from DataLoader, Fix is to set classpath with the directory where dataloader.bat, process.bat, and encrypt.bat are installed . Similar settings are used for the shell script (dataloader) in Mac for consistency.


ashitsalesforce:
closing without merge.

PR #669, opened 2023-04-22
replace softlink for linux with copy of util.sh and dataloader_consol…

ashitsalesforce:
…e from mac directory.

replace softlink for linux with copy of util.sh and dataloader_console from mac directory.


ashitsalesforce:
This PR enables CodeQL code scanning to complete successfully.

PR #706, opened 2023-06-26
Update dlbuilder.sh to use correct test skip parameter

jgleonard-takeda:
"skipTests" isn't applicable here, it needs to use "maven.test.skip"

### DIFF
--- a/dlbuilder.sh
+++ b/dlbuilder.sh
@@ -16,7 +16,7 @@ run_mvn() {
   zip -q -d ${HOME}/.m2/repository/org/apache/logging/log4j/log4j-core/${log4j_version_num}/log4j-core-${log4j_version_num}.jar org/apache/logging/log4j/core/appender/db/jdbc/JdbcAppender.class
 
   # build uber jar
-  mvn clean package -DskipTests 
+  mvn clean package -Dmaven.test.skip
 }
 
 #################


ashitsalesforce:
Hi @jgleonard-takeda , thanks for catching incorrect flag. While you are correct that `mvn clean package -Dmaven.test.skip` will skip tests, given that the intent is to run unit tests, the right command to execute is `mvn clean package`. Consequently, I am closing this PR without merge and will check in the fix in another PR.

jgleonard-takeda:
Hi @ashitsalesforce thanks, yes, I submitted this PR because it looked like the tests were purposely skipped.  Without it, there's a failure on a csv test.

[ERROR] Failures:
[ERROR]   CsvTest.testCSVReadUTF8BOMBasic:77->testCSVReadBasic:96

Can be reproduced with the following:
```
FROM alpine:latest as base
RUN apk update && apk --no-cache add bash openjdk17-jdk git maven zip
WORKDIR /tmp
RUN git clone https://github.com/forcedotcom/dataloader.git && \
    cd dataloader && \
    git checkout v58.0.3 && \
    git submodule init && \
    git submodule update
WORKDIR /tmp/dataloader

# run a mvn build to download dependencies from the central maven repo
RUN  mvn clean compile

# build uber jar
RUN mvn clean package
```

PR #715, opened 2023-07-05
Single Whitespaces are not replaced by None-breaking Whitespace

fmannreply:
Added nullsafe ahead and back look when escapeHTMLChars so only multiple whitespace occurences are replaced by non-breaking whitespace.

U+0020 = Normal Whitespace = ` `
U+00A0 = No-Break Space = `&nbsp;`

Recent behavior:
> This is     a scentence! <Github doesn't render the additional spaces here
> ThisU+0020isU+0020U+0020U+0020U+0020aU+0020scentence!

turns into
> ThisU+00A0isU+00A0U+00A0U+00A0U+00A0aU+00A0scentence!
Which break line breaks in Rich Text fields, as any words will be evaluated as one long word (Firefox & Chrome).

Expected behavior:
> This is     a scentence!
> ThisU+0020isU+0020U+0020U+0020U+0020aU+0020scentence!

turns into
> ThisU+0020isU+0020U+0020U+0020U+0020aU+0020scentence!
preserving single whitespaces so rich text fields break correctly.

I don't know if you accept PRs. The unit tests are still working and the code includes relevant comments, if there are any questions feel free to comment. :)

### DIFF
--- a/src/main/java/com/salesforce/dataloader/action/visitor/DAOLoadVisitor.java
+++ b/src/main/java/com/salesforce/dataloader/action/visitor/DAOLoadVisitor.java
@@ -324,7 +324,23 @@ public abstract class DAOLoadVisitor extends AbstractVisitor implements DAORowVi
         for (int i = 0, len = input.length(); i < len; i++) {
             char c = input.charAt(i);
             int cval = c;
-            if (Character.isWhitespace(c) || cval == NONBREAKING_SPACE_ASCII_VAL) {
+
+            char nextChar = 0;
+            if(i+1 < input.length()){
+                nextChar = input.charAt(i+1);
+            }
+            char prevChar = 0;
+            if(i>0){
+                prevChar = input.charAt(i-1);
+            }
+
+            boolean isCharWhitespace = Character.isWhitespace(c) || cval == NONBREAKING_SPACE_ASCII_VAL;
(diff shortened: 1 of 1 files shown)


ashitsalesforce:
Hi @fmannreply , 

Replacing U+0040 (&nbsp;) with U+0020 (&#32;) has the side effect that the browser displays only one space as you observed in your own comment: This is a scentence! <Github doesn't render the additional spaces here

Even though you had typed in 
```
This is     a scentence!
```

If we were to replace the very first whitespace character with U+0020 and subsequent characters with U+0040, we would see the following in the browser (one less whitespace character between 'is' and 'a'):
```
This is    a scentence!
``

fmannreply:
@ashitsalesforce yep, that's why only none-single whitespace are replaced with this code. If you have "SpaceSpace" both will be converted to none-breaking spaces.

If any is true: is first char, char before is whitespace, char after is whitespace, or is last char - therefore you end up displaying all spaces correctly.

If you just use all none-breaking whitespaces, the whole sentence is treated as a single word, therefore text doesn't break/wrap correctly.

ashitsalesforce:
@fmannreply , I removed `nextChar == 0` and `prevChar == 0` from the setting of booleans `isNextCharWhitespace` and `isPrevCharWhitespace` to avoid issue with a single whitespace setting both of them to true.

PR #772, opened 2023-08-21
fix to avoid issue #767

ashitsalesforce:
Users are confused by the need to check the checkbox for other delimiters and specify the delimiter values in Settings dialog. This fix is to remove the "Specify other delimiters checkbox" in Settings dialog. Specifying the delimiter values is sufficient to determine that other delimiter values are applicable.


ashitsalesforce:
The PR removes the "Allow other characters as CSV delimiters" checkbox because presence/absence of delimiters in "Other delimiters for load operations" text field is sufficient to determine if other characters are to be used as CSV delimiters.

PR #960, opened 2024-01-31
fix duplicate records and inconsistencies in the success log when using 1 batch size

mcgrinberg:


ashitsalesforce:
Closing without merge as an alternate fix is checked in - #986

PR #1353, opened 2024-10-22
fix broken link

HUMENTH:


ashitsalesforce:
Hi @HUMENTH , thanks for catching the issue and submitting a PR containing the fix. Given the installation documentation for both Mac and Windows is now on the same page, I will accept your PR and then modify the content to point to the same documentation.

PR #1435, opened 2025-01-05
Update install.command

fddemora:
updated path to source (linux/mac)

Only change was on line 4. The path /util/util.sh doesn't exist from the current directory.
The output after the change made (successful java version check, my environment doesn't have it yet):

`$ ./install.sh`
Data Loader requires Java JRE @@MIN_JAVA_VERSION@@ or later. Checking if it is installed...
Did not find java command.

Java JRE @@MIN_JAVA_VERSION@@ or later is not installed or DATALOADER_JAVA_HOME environment variable is not set.
For example, download and install Zulu JRE @@MIN_JAVA_VERSION@@ or later from here:
https://www.azul.com/downloads/

### DIFF
--- a/release/install/install.command
+++ b/release/install/install.command
@@ -1,7 +1,7 @@
 :; #!/bin/bash #
 :; #
 :; DL_INSTALL_ROOT="$(dirname "$(readlink -f "$0")")" #
-:; source "${DL_INSTALL_ROOT}/util/util.sh" #
+:; source "${DL_INSTALL_ROOT}/../../src/main/resources/mac/util/util.sh" #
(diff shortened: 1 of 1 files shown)


fddemora:
It says I signed the CLA.

ashitsalesforce:
Hi @fddemora , are you trying to run install.command from your local git clone's folder (`<your git clone root folder>/release/install` folder)? If so, that's not supported. The expected use of install.command is when it is located in the installation root folder with util.sh in util sub-folder. You need to build the zip artifact, extract its contents, go to the root folder of the extraction, and run install.command from that folder to use it per its intended use.

PR #1663, opened 2026-07-15
Fix issue 1647

shalomcarmel:
…environment
The headless CLI needs the SWT that is correct for the processor architecture, otherside it fails.


shalomcarmel:
CLA signed

